PR #76, opened 2014-12-24
Dynamically loadable HDF5 filter plugin

kiyo-masui:
HDF5 1.8.11+ supports "dynamically loadable filter plugins" (http://www.hdfgroup.org/HDF5/doc/Advanced/DynamicallyLoadedFilters/HDF5DynamicallyLoadedFilters.pdf). Following that recipe, I've made a plugin for the blosc filter.

I needed to do this anyway for my own work and thought I would put in a pull request upstream to see if there is interest in having this in blosc.  I'm sure this isn't up to standard for merging yet: you will probably want to change the format and content of the headers and I put in the bare minimum for the README.

### DIFF
--- a/hdf5/README.rst
+++ b/hdf5/README.rst
@@ -17,9 +17,15 @@ value is returned.
 An example C program ("example.c") is included which demonstrates the
 proper use of the filter.
 
+Alternatively, instead of registering the Blosc filter,  you can use the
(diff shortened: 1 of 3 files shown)


FrancescAlted:
Although I did not put a lot of effort in checking this, I think it is better to include it that having it waiting, so here we go.

PR #77, opened 2015-01-04
Only define __SSE2__ when compiling with MSVC

jack-pappas:
I've added preprocessor checks around the definition of the ``__SSE2__`` symbol on Windows. It's only necessary to define the symbol when compiling with MS Visual C++ (MSVC); other compilers that already define the symbol emit a warning because the symbol is being re-defined.

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -14,7 +14,14 @@
 #if defined(_WIN32) && !defined(__MINGW32__)
   #include <windows.h>
   #include "win32/stdint-windows.h"
-  #define __SSE2__          /* Windows does not define this by default */
+
+  /* Define the __SSE2__ symbol if compiling with Visual C++ and
+     targeting the minimum architecture level supporting SSE2.
+     Other compilers define this as expected and emit warnings
+     when it is re-defined. */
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Hmm, not an expert on MSVC, but the next:

https://social.msdn.microsoft.com/Forums/en-US/4b2d2f47-ae8a-487e-b9d4-494ced8b30c3/vs2010-vc-mix86fp-and-archavx

is suggesting:

#if defined( _M_X64 ) || _M_IX86_FP >= 1

(notice the 1 instead of 2).  Suggestions?

jack-pappas:
I made these changes based on this page: http://msdn.microsoft.com/en-us/library/b0084kay.aspx

The OP in the forum post you linked says they're detecting either SSE or SSE2, which is why they used ``_M_IX86_FP >= 1``. blosc needs SSE2 though, so my change here checks for ``_M_IX86_FP >= 2``.

I added a check for ``_M_IX86`` along with ``_M_IX86_FP`` because MSVC does support ARM targets as well (for phones and tablets) and ``__SSE2__`` shouldn't be defined in that case.

FrancescAlted:
Ok, this sounds good to me.

PR #78, opened 2015-01-04
Use the built-in stdint.h with MSVC when available

jack-pappas:
``stdint.h`` is available in MSVC 16.0 / VS2010 and newer. I've added preprocessor checks so the compatibility header is only imported when compiling with an earlier version of MSVC; otherwise, the built-in ``stdint.h`` is used.


FrancescAlted:
Look good.  Thanks!

PR #79, opened 2015-01-04
Modified examples to use the new context-based API

jack-pappas:
I've modified the blosc examples so they use the new context-based API functions instead of the older functions which entail use of the global context.

### DIFF
--- a/examples/many_compressors.c
+++ b/examples/many_compressors.c
@@ -53,10 +53,14 @@ int main(){
   static float data[SIZE];
   static float data_out[SIZE];
   static float data_dest[SIZE];
-  int isize = SIZE*sizeof(float), osize = SIZE*sizeof(float);
-  int dsize = SIZE*sizeof(float), csize;
-  int nthreads, pnthreads, i;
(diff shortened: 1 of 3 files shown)


FrancescAlted:
I think there is a misunderstanding here.  You call the current (and official) API 'old', but in the header docstrings it is never said that they are 'old' by no means.  In fact, they have interesting advanatges over the '_ctx' counterparts.  For details, see:

https://github.com/Blosc/c-blosc/pull/66#discussion_r20007507

I think it would be better if you can contribute a different PR with different examples using the '_ctx' API.

jack-pappas:
Thanks @FrancescAlted. I was under the impression that the non-ctx functions were being phased out in favor of the ctx-based functions. I'll read over the issue you linked to and perhaps submit a new PR which adds _ctx-based examples to the existing examples.

FrancescAlted:
One thing you can do is adding examples for the _ctx() API.  This API is meant mainly for multithreaded environments, so an example including a multithreaded app using c-blosc would be really cool.

PR #80, opened 2015-01-04
Use C11 aligned_alloc when it's available.

jack-pappas:
I've implemented a bit of code in ``blosc.c`` so when compiling with a compiler which supports C11 (which includes recent versions of gcc, clang, and icc), the built-in ``aligned_alloc`` function will be used to allocate aligned memory.


FrancescAlted:
Looks good.  Merging.

PR #81, opened 2015-01-04
Use llabs function instead of abs to avoid truncating the result

jack-pappas:
When compiling blosc with MSVC 18.0 (VS2013), I saw some compiler warnings about ``long long int`` values being possibly truncated by ``abs`` (because that function takes an ``int``). I've modified the code to use the ``llabs`` function instead to avoid this issue.


FrancescAlted:
True.  I had seen the warning also with recent clang (3.5?) too, so merging.

PR #86, opened 2015-02-17
SSE2 acceleration for shuffling larger types

jack-pappas:
I've implemented SSE2-accelerated shuffle/unshuffle functions which can be used when the type size is larger than the SSE2 vector type (i.e., >16 bytes).

I've also hand-unrolled all of the small inner loops within the existing shuffle/unshuffle routines (for 2/4/8/16-byte types). This could have been done by some compilers already (in theory, at least), but I ran a baseline benchmark, then benched again with the change and the performance improved noticeably. I've included the results below for comparison.

Baseline:

```
Blosc version: 1.5.3.dev ($Date:: 2014-12-30 #$)
List of supported compressors in this build: blosclz,lz4,lz4hc,zlib
Supported compression libraries:
  BloscLZ: 1.0.3
  LZ4: 1.5.0
  Snappy: unknown
  Zlib: 1.2.8
Using compressor: blosclz
Running suite: suite


--> 1, 536870912, 16, 19, blosclz, 1
********************** Run info ******************************
Blosc version: 1.5.3.dev ($Date:: 2014-12-30 #$)
Number of threads: 1
Using synthetic data with 19 significant bits (out of 32)
Working set: 512.0 MB
Dataset size: 536870912 bytes
Type size: 16 bytes
Element Count: 33554432
********************** Running benchmarks *********************
memcpy(write):		 47219.8 us, 10842.9 MB/s
memcpy(read):		 88055.2 us, 5814.5 MB/s
Compression level: 0
comp(write):	 58532.1 us, 8747.3 MB/s	  Final bytes: 536870928  Ratio: 1.00
decomp(read):	 50988.6 us, 10041.5 MB/s	  OK
Compression level: 1
comp(write):	 214390.2 us, 2388.2 MB/s	  Final bytes: 77791248  Ratio: 6.90
decomp(read):	 108376.0 us, 4724.3 MB/s	  OK
Compression level: 2
comp(write):	 230685.9 us, 2219.5 MB/s	  Final bytes: 45645840  Ratio: 11.76
decomp(read):	 118848.3 us, 4308.0 MB/s	  OK
Compression level: 3
comp(write):	 236344.4 us, 2166.3 MB/s	  Final bytes: 45645840  Ratio: 11.76
decomp(read):	 118475.5 us, 4321.6 MB/s	  OK
Compression level: 4
comp(write):	 270834.0 us, 1890.5 MB/s	  Final bytes: 40583184  Ratio: 13.23
[...]

PR #88, opened 2015-02-22
SSE2 shuffle/unshuffle functions for large types

jack-pappas:
This PR is a more-focused version of #86. It includes changes to add the SSE2-accelerated shuffle/unshuffle functions for types larger than the vector size (in this case, 16 bytes). The changes also allow the SSE2-accelerated functions to be used in some cases where the input data doesn't contain an "ideal" number of elements. In those cases, the number of elements is rounded down to an acceptable value and the SSE2-accelerated functions are used (which generally covers the majority of the elements in the input buffer); the simple byte-by-byte shuffle/unshuffle functions are used for the remaining elements. This is particularly important for "large" types: due to the way block sizes are calculated (prior to these changes), data shuffled with existing/older versions of blosc won't have the correct size to utilize the new SSE2-accelerated functions -- so these modifications allow the new functions to be used with existing shuffled/compressed data.

I'm opening the PR now to get feedback, but I think additional tests need to be implemented to verify data is round-tripped correctly and is backwards-compatible with the existing code (the byte-by-byte shuffle/unshuffle for large types) before this is merged. I'll implement some tests for this ASAP, but I encourage others to test the code as well (in whatever way is appropriate for their usage of blosc) to help sniff out potential edge cases.

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -1115,8 +1115,8 @@ void unshuffle(size_t bytesoftype, size_t blocksize,
       unshuffle_tiled(_dest, _src, vectorizable_bytes, bytesoftype);
     }
     else {
-      /* Non-optimized shuffle */
-      _shuffle(bytesoftype, blocksize, _src, _dest);
+      /* Non-optimized unshuffle */
+      _unshuffle(bytesoftype, blocksize, _src, _dest);
       /* The non-optimized function covers the whole buffer,
          so we're done processing here. */
       return;
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -1017,7 +1017,7 @@ unshuffle_tiled(uint8_t* dest, const uint8_t* orig, size_t size, size_t bytesoft
       int j;
 
       /* Load the first 128 bytes in 16 XMM registers */
-      uint8_t* const src_for_ith_element = orig + i;
+      const uint8_t* const src_for_ith_element = orig + i;
       for (j = 0; j < 16; j++) {
         xmm1[j] = _mm_loadu_si128((__m128i*)(src_for_ith_element + (num_elements * (offset_into_type + j))));
       }
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -76,11 +76,11 @@ shuffle2_AVX2(uint8_t* dest, const uint8_t* src, size_t size)
   size_t i, j, k;
   size_t nitem;
   __m256i a[2], b[2], c[2], d[2], shmask;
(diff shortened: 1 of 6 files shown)


FrancescAlted:
Thanks for decoupling this feature in a proper PR.  Any idea on why TravisCI is complaining on:

```
Test  #1: test_test_basics .................***Exception: Numerical  0.00 sec
```

?

jack-pappas:
Yes -- it looks to be due to the change on line 887 (in my version). The case where ``typesize`` is 1 and ``nbytes`` is 7 causes the function to return 0 for the block size, which results in division by zero in ``initialize_context_compression``. I'm working to fix the bug now.

jack-pappas:
I've fixed the broken test in test_basics. I've also added some additional type sizes to test with in #90. The new tests there all pass when run against the current ``master``; I see a test failure when running them against the code in this PR though, so I'll continue bug-hunting.

jack-pappas:
Ok, I've fixed the bit of code that was breaking the new tests in #90, and they all pass now. I'm going to implement some additional tests (within the next few days) to verify the new code in this PR can read data shuffled/compressed by the current code and vice versa.

blosc/shuffle.c
```diff
@@ -312,12 +393,11 @@ static void
...
 {
   size_t i, j, k;
-  size_t neblock, numof16belem;
-  __m128i xmm0[4], xmm1[4];
+  const size_t numof16belem = size / (4 * sizeof(__m128i));
 
-  neblock = size / 4;
-  numof16belem = neblock / 16;
   for (i = 0, k = 0; i < numof16belem; i++, k += 4) {
+    __m128i xmm0[4], xmm1[4];
```
FrancescAlted: Is this an example of intermingled declarations and code that MSVC would have problems with, or is this supported?

FrancescAlted:
This is looking good so far.  Could you please post some small benchmark on speed improvements for typesizes larger than 16 bytes?  I am a curious guy :)

jack-pappas:
I can post benchmarks sometime this week. However, please wait to merge this patch - I have since found that it exposes an existing bug in blosc. I'll contact you privately to discuss the issue.

FrancescAlted:
Ah, ok.  I will be glad to discuss the problem.

FrancescAlted:
Hey Jack, still waiting for discussing this existing bug in c-blosc that you mentioned here time ago :)

jack-pappas:
I've figured out the bug, and it was a subtle issue hiding in my changes, not in blosc. I'm closing this issue as it's partially superseded by #129. Once those changes are merged in, I'll submit a new PR with the remaining changes from this PR (the large type support).

PR #90, opened 2015-02-22
Added additional type sizes to test_basics

jack-pappas:
This change adds additional type sizes to the ``test_shuffle``, ``test_noshuffle``, and ``test_getitem`` tests in ``test_basics.c``. These type sizes are (mostly) larger than the SSE2 vector size of 16 bytes in order to test how larger types are handled.


FrancescAlted:
Merged.  Thanks.

PR #91, opened 2015-02-22
Use memcmp to check round-trip results in tests

jack-pappas:
This PR modifies ``test_basics.c`` so it uses the ``memcmp`` function to verify round-tripped data matches the original. ``memcmp`` is generally faster than the byte-by-byte comparison so it speeds up execution of the tests.


FrancescAlted:
Merged.  Thanks!

PR #94, opened 2015-03-09
do not use win32/pthreads code with win32 gcc (v1.4 based)

ARF1:
Tested with win32 GCC & mingw32:

win32 GCC defines `__GNUC__` but does not need `pthread.c`
win32 mingw32 defines `__GNUC__` & `__MINGW32__` and does need `pthread.c`

Other compilers:
non-gcc compilers define `_WIN32` and probably need `pthread.c`

other gcc based win32 compilers untested

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -42,7 +42,7 @@
   #include <inttypes.h>
 #endif  /* _WIN32 */
 
-#if defined(_WIN32)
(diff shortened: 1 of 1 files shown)


ARF1:
This was a doomed attempt of using preprocessor definitions to distinguish between "TDM-GCC" and "pure" MinGW due to an confusion about the nature or "TDM-GCC".

Note: c-blosc is incompatible with TDM-GCC due to the inclusion of a pthread emulation library that is not stable with c-blosc.

PR #98, opened 2015-03-30
AVX2 version shuffle and unshuffle

littlezhou:
The first commit is the implementation of AVX2 functions  and some changes have been made in order to make it work properly. The second commit adds a standalone testing, it compares the the time need to shuffle/unshuffle for AVX2 and SSE2.

PR #99, opened 2015-04-14
AVX2 version of shuffle functions added.

littlezhou:


FrancescAlted:
Supersedes #98

FrancescAlted:
It is a pity that I don't have any box reachable with AVX2.  @esc maybe you have one AVX2 machine so that you can give this a try and report your experience?

blosc/blosc.c
```diff
@@ -198,6 +199,26 @@ static uint8_t *my_malloc(size_t size)
...
+  /* Mac OS X guarantees 16-byte alignment in small allocs */
+  posix_memalign( &block, 32, size);
+#elif _POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600
+  /* Platform does have an implementation of posix_memalign */
+  res = posix_memalign(&block, 32, size);
+#else
+  block = malloc(size);
+#endif  /* _WIN32 */
+
+#else  /* else __AVX2__ */
```
FrancescAlted: Hmm, I am wondering if it would not be better to try to align to 32 bytes by default and reduce code complexity.  Do you see any inconvenient in doing this?

littlezhou:
It's great, i don't like these preprocessor directives either.

PR #100, opened 2015-04-19
Fix preprocessor macros in shuffle.c that broke MSVC compilation

jack-pappas:
The recent addition of AVX2 support in shuffle.c broke compilation with MSVC due to an "invalid integer expression" on line 530 (``#elif __SSE2__``). I've fixed this so the preprocessor checks for the definition of the symbol rather than the value, and I've modified the ``__AVX2__`` check to be consistent (so it also checks whether the symbol is defined).

I've also made a small modification to the preprocessor checks which define the ``__SSE2__`` symbol for MSVC; now it only defines the symbol if it's not already defined. This avoids symbol redefinition warnings when using other compilers (e.g., clang) in Visual Studio.

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -19,7 +19,8 @@
      targeting the minimum architecture level supporting SSE2.
      Other compilers define this as expected and emit warnings
      when it is re-defined. */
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Looks good.  Thanks!

Issue #101, opened 2015-04-20
AVX2 binaries won't be able to re-distribute


FrancescAlted:
@juliantaylor just warned about the possibility that if someone creates a binary with AVX2, then this won't work in non-SSE2 machines.  This is in fact a big problem because Blosc tends to be embedded in larger binary packages.  The only solution would be to disable AVX2 usage on non-AVX2 machines.  @littlezhou do you think this can be done easily?

juliantaylor:
the best way to get the best performance from the available hardware is to dispatch to different functions depending on the hardware at runtime, e.g. via GCC's `__builtin_cpu_supports("avx")` though doing that compiler independent is not as nice. I can probably give this a try at least for gcc only.

juliantaylor:
see #102 for how one could do it

FrancescAlted:
PR #108 also addresses this in a more general way.  Closing.

PR #102, opened 2015-04-20
use runtime detection of AVX2

juliantaylor:
allows building portable binaries that still can use AVX2 on appropriate
hardware.
Currently only supported compiler is GCC >= 4.8 (avx2 support landed in 4.7)
and increases alignment requirement to 32 for all vectorized shuffles.
This could be relaxed with more code or even removed, to my knowledge
haswell, unlike sandy bridge, does not suffer any measurable penalty for
not 32 byte aligned memory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,11 +170,6 @@ endif (CMAKE_C_COMPILER_ID STREQUAL GNU)
 # Set the "-msse2" build flag only if the CMAKE_C_FLAGS is not already set.
 # Probably "-msse2" should be appended to CMAKE_C_FLAGS_RELEASE.
 if(CMAKE_C_COMPILER_ID STREQUAL GNU OR CMAKE_C_COMPILER_ID STREQUAL Clang OR CMAKE_C_COMPILER_ID STREQUAL Intel)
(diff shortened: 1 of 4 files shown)


FrancescAlted:
Looks very good.  Thanks a  lot!

FrancescAlted:
I have tested that compiling on an avx2 machine and copying the executable into a non-avx2 machine works well, at least for GCC/Linux, which is great.

@jack-pappas would you like to try with a MSVC compiler?  I suppose you will need to give support for it in CMakeLists.txt.

juliantaylor:
this will still need a followup for the alignment, as default alignment is 16 bytes this will cause some input to not even use SSE2 when it could. I can probably do that in the next few days.

FrancescAlted:
You mean here https://github.com/Blosc/c-blosc/blob/master/blosc/shuffle.c#L283, right?  Yep, that should be addressed.  Will open a ticket so as to not forget this.

FrancescAlted:
It is in #103

jack-pappas:
@FrancescAlted I've added the necessary support for MSVC in #108. That PR also handles #101.

Issue #103, opened 2015-04-21
Better check for 16 bytes alignment


FrancescAlted:
In https://github.com/Blosc/c-blosc/blob/master/blosc/shuffle.c#L283 one makes a check for 32-bytes alignment and if this is not fulfilled, then it falls backs to non-SIMD code.  One should add at least an additional check for 16 bytes and in that case, call the SSE2 code.

PR #104, opened 2015-04-21
remove 32 byte alignment requirement by using unaligned instructions

juliantaylor:
loading 32 byte aligned data has no penalty on haswells which is
required for AVX anyway.
Loading 16 byte aligned data has a small (~5%-10%) penalty but the old
code would have fallen back to SSE2 which is even slower so this is
still a net improvement.
Closes gh-103

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -280,8 +280,10 @@ shuffle16_SSE2(uint8_t* dest, const uint8_t* src, size_t size)
 /* Shuffle a block.  This can never fail. */
 void shuffle(size_t bytesoftype, size_t blocksize,
              const uint8_t* _src, uint8_t* _dest) {
-  int unaligned_dest = (int)((uintptr_t)_dest % 32);
-  int multiple_of_block = (blocksize % (32 * bytesoftype)) == 0;
+  int unaligned_dest = (int)((uintptr_t)_dest % 16);
+  int multiple_of_block = have_avx2() ?
+      (blocksize % (32 * bytesoftype)) == 0 :
+      (blocksize % (16 * bytesoftype)) == 0;
   int too_small = (blocksize < 256);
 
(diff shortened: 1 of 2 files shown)


blosc/shuffle.c
```diff
@@ -529,9 +531,11 @@ void unshuffle(size_t bytesoftype, size_t blocksize,
 /* Unshuffle a block.  This can never fail. */
 void unshuffle(size_t bytesoftype, size_t blocksize,
                const uint8_t* _src, uint8_t* _dest) {
-  int unaligned_src = (int)((uintptr_t)_src % 32);
-  int unaligned_dest = (int)((uintptr_t)_dest % 32);
-  int multiple_of_block = (blocksize % (32 * bytesoftype)) == 0;
+  int unaligned_src = (int)((uintptr_t)_src % 16);
```
FrancescAlted: Hmm, I am tempted to use the 'storeu' functions for the SSE2 flavor as well, and then we can completely relax the alignment condition.  But I suppose that if I used the direct assignment in first place is because the difference is speed can be quite important.  But worth another experiment anyway :)

FrancescAlted:
Oh, that's very nice.  Thanks!

FrancescAlted:
Just as a reference, my benchmarks say that compression speed is mostly unaffected by this, but that decompression speed is 10% slower when using 1 thread, but that is reduced to a 4% when using 8 threads (using a Intel Xeon E3-1240 v3 @ 3.40GHz).  Well, still AVX2 has *some* advantage anyway.

juliantaylor:
hm a Xeon E3-1240 does not have avx2 support so it should be completely unaffected by this change

juliantaylor:
oh forgot to add v3 to the google query
10% slower is not good, I could not see a difference on an i5, I can recheck on a E5-4620 later.

PR #105, opened 2015-04-22
Removed the alignment requirement for using SSE2 or AVX2 code

FrancescAlted:
Implemented completely unaligned loads and stores for SSE2 code.  As AVX2 code is already supporting unaligned loads and stores, this PR is completely removing the need for buffer alignment so as to use SSE2 or AVX2 instructions.

Benchmarks seems to indicate that the use of 'loadu' and 'storeu' initrinsics is negligible, so I think that is good for merging, unless @juliantaylor or @littlezhou would pose objections to this (I hope not).

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -350,8 +350,8 @@ unshuffle2_SSE2(uint8_t* dest, const uint8_t* orig, size_t size)
   numof16belem = neblock / 16;
   for (i = 0, k = 0; i < numof16belem; i++, k += 2) {
     /* Load the first 32 bytes in 2 XMM registrers */
-    xmm1[0] = ((__m128i *)orig)[0*numof16belem+i];
-    xmm1[1] = ((__m128i *)orig)[1*numof16belem+i];
+    xmm1[0] = _mm_loadu_si128((__m128i*)(orig) + 0*numof16belem+i);
+    xmm1[1] = _mm_loadu_si128((__m128i*)(orig) + 1*numof16belem+i);
     /* Shuffle bytes */
     /* Compute the low 32 bytes */
     xmm2[0] = _mm_unpacklo_epi8(xmm1[0], xmm1[1]);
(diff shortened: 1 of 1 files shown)


juliantaylor:
the intel core duo and older AMDs do still have penalties for unaligned load instructions used on aligned data, so that could slow things down.
default malloc alignment on 64 bit machines is 16 bytes so most input buffers should already be aligned.

One could also adapt the shuffle code itself to do a few peeling iterations to at least get one of the buffers aligned. But I don't think its worth the extra code complexcity

FrancescAlted:
Yes, having in account 'old' machines is a good point, and you are right that most of the buffers out there are already aligned to 16 bytes.

Enviat des del meu iPhone

El 22/04/2015, a les 19:08, Julian Taylor <notifications@github.com> va escriure:

jack-pappas:
I think this is a good change. If the unaligned load penalty on older hardware becomes a show-stopping issue for some blosc users and you want to keep supporting that hardware, it should be fairly straightforward to extend the new CPU feature-detection code in 1.6.0 to determine whether the CPU supports efficient misaligned loads and if not, dispatch to a special version of the SSE2 shuffle (which would actually be the current version, before this PR).

juliantaylor:
I don't think this change is worth it, we make it slower on some older machines and gain almost nothing in return, data is already 16 byte aligned in pretty much all cases.

I'm still using an amd phenom and an intel core2 and will continue to do so until they break, both have small penalties.

FrancescAlted:
While I agree with @juliantaylor , there is still the case of 32-bit platforms where alignment by default seems to be of 8 byte (see e.g. https://msdn.microsoft.com/en-us/library/ycsb6wwf.aspx), and this patch actually helps in using SSE2 in these scenarios.

There is still the issue of how much the performance is affected for old processors, and I would like to see some figures before taking a decision in this regard.

FrancescAlted:
Overridden byt #130

PR #108, opened 2015-04-25
Implement CPU feature detection and dispatch

jack-pappas:
I've implemented the CPU feature detection (for x86/x64) proposed in #101. These changes make it possible to compile one binary that runs on a system that supports SSE2 or AVX2 (or neither), so the redistribution problem is fixed.

As part of the change, I've split the AVX2 and SSE2 shuffle implementation out into separate files. The new setup makes it easier to add additional hardware support going forward.

I've removed the host-CPU-feature detection from the CMake build system, as it no longer makes sense -- it doesn't matter which CPU features the host supports, it only matters what the compiler supports (and which hardware-specific implementations blosc includes). In other words, you want to be able to build binaries targeting new hardware on old hardware.

I've tested these changes with VS 2013 on Windows 7 on a Haswell CPU and they seem to work as expected (and the tests pass).

Some small additional details about the changes can be found in the commit messages or in the comments in the code.


blosc/shuffle.c
```diff
@@ -7,585 +7,252 @@
...
-    }
-  }
-  leftover = blocksize % bytesoftype;
-  memcpy(_dest+neblock*bytesoftype, _src+neblock*bytesoftype, leftover);
-}
-
+/*  Include hardware-accelerated shuffle/unshuffle routines based on
+    the target architecture. Note that a target architecture may support
+    more than one type of acceleration!*/
+#if defined(__AVX2__)
```
juliantaylor: this must not be defined in the portable file as it implies  the compiler has been allowed to use any ISA up to avx2 for the file

blosc/shuffle.c
```diff
@@ -7,585 +7,252 @@
...
+#endif
+void init_shuffle_implementation() {
+  /* Initialization could (in rare cases) take place concurrently on multiple threads,
+     but it shouldn't matter because the initialization should return the same result
+     on each thread (so the implementation will be the same). Since that's the case
+     we can avoid complicated synchronization here and get a small performance benefit
+     because we don't need to perform a volatile load on the initialization variable
+     each time this function is called. */
+#if defined(__GNUC__) || defined(__clang__)
+  if (__builtin_likely(!implementation_initialized, 0)) {
```
juliantaylor: gcc has no __builtin_likely, you probably want __builtin_expect

blosc/shuffle.c
```diff
@@ -7,585 +7,252 @@
...
+    /*  Set the flag indicating the implementation has been initialized.
+        Use an atomic exchange if supported; it's not strictly necessary
+        but helps avoid unnecessary re-initialization. */
+#if __STDC_VERSION__ >= 201112L
+    atomic_exchange(&implementation_initialized, 1);
+#elif defined(_MSC_VER)
+    _InterlockedExchange(&implementation_initialized, 1);
+#else
+    implementation_initialized = 1;
+#endif
```
juliantaylor: this doesn't actually prevent re-initialization, just disallows compiler reordering

juliantaylor:
the reorganization makes sense, though it would be nice too keep using the simple __builtin_cpu_supports for gcc >= 4.7. It also boils down to a value lookup from a table that is initialized during library load.

blosc/shuffle-avx2.h
```diff
@@ -0,0 +1,41 @@
...
+#include "shuffle-common.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+  AVX2-accelerated shuffle routine.
+*/
+DLL_EXPORT void shuffle_avx2(const size_t bytesoftype, const size_t blocksize,
```
juliantaylor: definition of DLL_EXPORT is not included in this file, also stdint.h is needed

juliantaylor:
it doesn't seem to work on my linux haswell and amd phenom, it chooses the generic shuffle on both

jack-pappas:
@juliantaylor Thank you for the feedback -- you spotted several subtle issues, which I've now gone back and fixed. I believe the reason you got the generic shuffle on your Linux Haswell and AMD Phenom is that GCC doesn't properly support the ``_xgetbv`` intrinsic provided by MSVC and ICC; in GCC it's defined but always returns zero. I added a bit of code to the CPU feature detection function to dump out all of the values then saw that SSE2 and AVX2 were "available" but weren't shown as having state-save OS support. I implemented ``_xgetbv`` for those compilers and now it works as expected (I confirmed with Linux on Haswell and Linux in a VirtualBox VM (which supports only SSE2)).

@FrancescAlted The CI build is passing now (I rebased this PR after you merged #109).

blosc/shuffle.c
```diff
@@ -7,585 +7,278 @@
...
+    "=D" (cpuInfo[1]),
+#else
+    "cpuid":
+    "=b" (cpuInfo[1]),
+#endif  /* defined(__i386) && defined(__PIC__) */
+    "=a" (cpuInfo[0]),
+    "=c" (cpuInfo[2]),
+    "=d" (cpuInfo[3]) :
+    "a" (function_id), "c" (subfunction_id)
+    );
```
FrancescAlted: First time that I will be adding pure assembler into blosc.  I can do that, but really, there is not another way to do that for GCC/clang?

blosc/shuffle.c
```diff
@@ -7,585 +7,278 @@
...
+    /* "xgetbv"
+       This is specified as raw instruction bytes due to some older compilers
+       having issues with the mnemonic form.
+    */
+    ".byte 0x0f, 0x01, 0xd0":
+    "=a" (eax),
+    "=d" (edx) :
+    "c" (xcr)
+    );
+  return ((uint64_t)edx << 32) | eax;
```
FrancescAlted: same here

blosc/shuffle-generic.c
```diff
@@ -0,0 +1,43 @@
...
+
+  /* Non-optimized shuffle */
+  neblock = blocksize / bytesoftype;  /* Number of elements in a block */
+  for (j = 0; j < bytesoftype; j++) {
+    for (i = 0; i < neblock; i++) {
+      _dest[j*neblock+i] = _src[i*bytesoftype+j];
+    }
+  }
+  leftover = blocksize % bytesoftype;
+  memcpy(_dest + neblock*bytesoftype, _src + neblock*bytesoftype, leftover);
```
FrancescAlted: I am getting this warning with GCC:

```
/home/faltet/blosc/jack-pappas-c-cblosc/blosc/shuffle-generic.c: In function ‘shuffle_generic’:
/home/faltet/blosc/jack-pappas-c-cblosc/blosc/shuffle-generic.c:25:3: warning: incompatible implicit declaration of built-in function ‘memcpy’
   memcpy(_dest + neblock*bytesoftype, _src + neblock*bytesoftype, leftover);
```

blosc/shuffle-generic.c
```diff
@@ -0,0 +1,43 @@
...
+
+  /* Non-optimized unshuffle */
+  neblock = blocksize / bytesoftype;  /* Number of elements in a block */
+  for (i = 0; i < neblock; i++) {
+    for (j = 0; j < bytesoftype; j++) {
+      _dest[i*bytesoftype+j] = _src[j*neblock+i];
+    }
+  }
+  leftover = blocksize % bytesoftype;
+  memcpy(_dest+neblock*bytesoftype, _src+neblock*bytesoftype, leftover);
```
FrancescAlted: And here too:

```
/home/faltet/blosc/jack-pappas-c-cblosc/blosc/shuffle-generic.c: In function ‘unshuffle_generic’:
/home/faltet/blosc/jack-pappas-c-cblosc/blosc/shuffle-generic.c:42:3: warning: incompatible implicit declaration of built-in function ‘memcpy’
   memcpy(_dest+neblock*bytesoftype, _src+neblock*bytesoftype, leftover);
```

FrancescAlted:
I really like this PR, and my benchmarks are showing that the SSE2 and AVX2 a properly recognized on Linux, so I am going to merge and fix the small pending warnings later on.  Thank you Jack!

blosc/shuffle.c
juliantaylor: for gcc you can use the intrinsic that was used before this PR, though there is probably no way around asm on windows. For linux in general you probably somethimes have `cpuid.h` but I have never tested that with clang.

juliantaylor:
on linux now the library is exporting functions it probably shouldn't.

```
nm blosc/libblosc.so
...
0000000000005df0 t shuffle16_avx2
0000000000004da0 t shuffle16_sse2
00000000000059d0 t shuffle8_avx2
0000000000006340 T shuffle_avx2
0000000000004c80 T shuffle_generic
0000000000005220 T shuffle_sse2
0000000000006ab0 T unshuffle
0000000000006070 t unshuffle16_avx2
0000000000004fd0 t unshuffle16_sse2
0000000000005bf0 t unshuffle8_avx2
0000000000006620 T unshuffle_avx2
0000000000004d10 T unshuffle_generic
0000000000005650 T unshuffle_sse2
```

juliantaylor:
SSE2 detection is also still broken for me

juliantaylor:
test_basics is now failing when built with clang, not sure yet why

juliantaylor:
fun the clang failure looks like a compiler bug, clang mixes up the argument order ...

PR #109, opened 2015-04-25
Travis-CI CMake fix

jack-pappas:
This is a fix to upgrade the CMake version used on Linux-based Travis CI builds. The currently supported version is 2.8.7, but #108 requires CMake 2.8.10.

This is a temporary change, as Travis-CI is currently preparing to roll out new Ubuntu 14.04LTS-based images which will have CMake 2.8.12 by default.

I've also enabled OS X workers for the Travis-CI builds for better coverage.


FrancescAlted:
OK, looks good to me.

PR #111, opened 2015-04-26
Fix bench program crash with MSVC

jack-pappas:
The change in 95a784c2 which defined ``posix_memalign`` for ``_WIN32`` / MSVC builds causes the ``bench`` tool to crash because buffers allocated with ``_aligned_alloc()`` need to be freed with ``_aligned_free()`` instead of the usual ``free()``. This PR defines a ``aligned_free`` macro which calls the appropriate function to free the aligned memory.


FrancescAlted:
Looks good to me.

PR #107, opened 2015-04-23
Pthreads for Windows support (v1.6.0 based)

yasushima-gd:
Fix #92.
Tested with `MinGW` and `TDM-GCC `.

`TDM-GCC`
```
>gcc --version
gcc (tdm64-1) 4.9.2
```
`MinGW`
```
>gcc --version
gcc (i686-posix-dwarf-rev2, Built by MinGW-W64 project) 4.9.1
```

PR #112, opened 2015-04-27
simd detection refactoring

juliantaylor:
split @jack-pappas cpuid and shuffle dispatch code into two functions  to make extension with other functions a bit simpler
add use of gcc's __builtin_cpu_supports back as it may be a bit more robust to other platforms than the other low level code
An issue with this is that the code for non-gcc might bitrot due to not being used anymore by a large fraction of users.

Also remove the xmm state check as I don't think an OS without that but with sse2 cpu support is relevant. This fixes the SSE2 detection for the clang compiler.

Also hide the shuffle symbols from the DSO, `write_compression_header` and `initialize_context_compression` are still public but removing them would technically be an ABI break.

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -14,6 +14,11 @@
 #include <stdio.h>
 #include <string.h>
 
+#if !defined(__clang__) && defined(__GNUC__) && defined(__GNUC_MINOR__) && \
+    __GNUC__ >= 5 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 8)
+#define HAVE_CPU_FEAT_INTRIN
+#endif
+
 /*  Include hardware-accelerated shuffle/unshuffle routines based on
     the target architecture. Note that a target architecture may support
     more than one type of acceleration!*/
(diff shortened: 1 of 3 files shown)


juliantaylor:
the clang miscompile should probably also be handled in the dispatch code to avoid wrong results. I'll follow that up when I have an answer from llvm devs

FrancescAlted:
After merging #107 you will need to rebase this, sorry.

juliantaylor:
done

jack-pappas:
@juliantaylor Why do you think GCC's ``__builtin_cpu_supports()`` would be more robust than what I wrote? The new CPU feature detection in blosc should cover all platforms and CPU vendors; why not just leave it as-is until we have reason to believe otherwise?

GCC being open-source, you should be able to track down their implementation (probably either in libgcc or injected by the compiler) and compare it to ours to see if it's substantially different; if it is, *and* it's going to produce substantially better results, then we should consider adding a special case for GCC to use ``__builtin_cpu_supports()``.

Can you provide more details about the SSE2 state detection failing with Clang? I think you're right that it *probably* won't matter, but why not just skip the check when compiling with Clang (i.e., so the check still exists for other compilers) instead of removing it altogether?

Hiding the extra symbols looks like a good change.

juliantaylor:
I believe its more robust because it works while yours needed changes, the xmm state detection won't work on older cpus (because there is no xsave) and _XCR_XFEATURE_ENABLED_MASK is not defined with gcc or clang

you could argue that now its probably good enough, but why risk it?
that issues in the less used code might get discovered later than they could be is a concern, but I think having no (probably hypothetical) bugs or compilation failures in the other fraction of users is an advantage too.

jack-pappas:
I defined ``_XCR_XFEATURE_ENABLED_MASK`` in the code so it is available for GCC and Clang; it's just defined to be ``0`` -- the same value it's defined to by MSVC and ICC: https://software.intel.com/sites/products/documentation/doclib/iss/2013/compiler/cpp-lin/GUID-6DBD0CC1-884D-40B9-8AA3-E568231ABE4E.htm

Digging through the Intel developer docs, I realize I made a small error in the XMM / YMM state detection based on the result of _xgetbv(). AVX2 is supported if bits 1 and 2 are set (for XMM and YMM state restore), but I made the assumption that bit one being set was required for XMM state restoration when it's not -- older CPUs that don't have ``xstore`` will use ``fxstore`` instead to restore the FPU and XMM state. So getting rid of the check like you did was correct and should fix the code so it works on older CPUs as well.

I'm still strongly in support of having one CPU detection method that covers everything. Especially now that I know why the SSE2 state detection didn't work in some cases (because it wasn't meant to be used that way, i.e., when only SSE2 is available and not AVX/AVX2).

blosc/shuffle.c
```diff
@@ -171,8 +198,32 @@ get_shuffle_implementation() {
...
+  #pragma message("Hardware-acceleration detection not implemented for the target architecture. Only the generic shuffle/unshuffle routines will be available.")
+  #else
+  #warning Hardware-acceleration detection not implemented for the target architecture. Only the generic shuffle/unshuffle routines will be available.
+  #endif
+
+#endif
+
+static shuffle_implementation_t
+get_shuffle_implementation() {
+  blosc_cpu_features f = blosc_get_cpu_features();
```
FrancescAlted: I prefer more meaningful function names than just `f`, but I can choose a name myself (like `get_cpu_features`?)

FrancescAlted:
This looks very good to me, so merging.  I agree with @jack-pappas that having one CPU detection method that covers everything would be desirable, but this is much better that what we had before.  Thanks!

PR #116, opened 2015-05-07
fix build on non-intel architectures

keszybz:
Build on arm was failing. First, the test for sse2 was detecting it as an intel processor, passing -msse2 to the compiler which cannot work. Second, link was failing because of a missing function.

Build with those patches: https://kojipkgs.fedoraproject.org//work/tasks/9975/9679975/build.log


FrancescAlted:
Looks good.  Thanks!

FrancescAlted:
BTW, any reason you don't include blosclz in your tests?

keszybz:
No idea, that that is quite old — was blosclz always there?

Anyway, here's a build with blosclz: https://kojipkgs.fedoraproject.org//work/tasks/831/9680831/build.log.

PR #119, opened 2015-05-10
Build with CMake

jack-pappas:
CMake can also be used to run the build, via ``cmake --build``. Building this way has the advantage of being cross-platform (the command is the same across all platforms) and independent of the build system being used (e.g., ``make``, ``msbuild``, etc.).


FrancescAlted:
Nice one.

PR #120, opened 2015-05-10
Fix platform detection on Windows 64-bit

jack-pappas:
This fixes an issue introduced by #116 where the processor type isn't detected correctly when building on Windows. On Windows, CMake specifies the system type as ``AMD64`` but on Linux specifies it as ``amd64``; the comparison is case-sensitive so it failed.


FrancescAlted:
Hmm, tricky.  Thanks!

PR #122, opened 2015-05-10
Added EditorConfig spec to help normalize code formatting

jack-pappas:
I've added an [``.editorconfig``](http://editorconfig.org/) spec to the repository to help keep the formatting of the code (e.g., indentation) consistent. The presence of this file doesn't *force* the formatting of the code to be correct, but many tools now support EditorConfig and will automatically pick up formatting settings from it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+; Top-most EditorConfig file
+root = true
+
+; Global settings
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Pretty cool!  Thanks!

PR #114, opened 2015-04-29
Implemented shuffle/unshuffle roundtrip tests

jack-pappas:
I've extended ``test_basics`` to add some roundtrip tests between the generic and accelerated shuffle/unshuffle routines to check that they produce the same results (e.g., that data shuffled with SSE2 could be unshuffled with the generic or AVX2 implementations).

These tests are fine to merge as-is, but the AVX2 round-trip tests are currently disabled (in ``tests/CMakeLists.txt``) until we figure out a better way to do the runtime CPU feature detection there. I suppose we could just copy it over from ``shuffle.c`` but that seemed kind of gross. For now, we'll at least have the generic <-> SSE2 tests to provide some coverage.

PR #123, opened 2015-05-10
Hidden symbols fix

jack-pappas:
This PR modifies the symbol hiding introduced in #112 and should allow #114 to be re-submitted without breaking the build.

When ``BUILD_TESTS`` is enabled in the CMake configuration, we now build a second, special version of ``libblosc``/``blosc.dll`` that exposes the normally-hidden symbols like ``shuffle_sse2`` to allow the tests to call them directly without going through the CPU feature-detection code. This extra library is called ``libblosc_testing``/``blosc_testing.dll`` and is *not* installed like the other libraries -- it is only used by and for the tests.

One thing that should be checked before accepting this PR -- the CMake config script now sets the symbol visibility to hidden by default (e.g., it will pass ``-fvisibility=hidden`` to GCC and Clang), so this might end up hiding some additional symbols that were visible by default before. If that's the case *and* they still need to be exported (for backwards compatibility), please let me know so I can make the necessary adjustments to preserve their visibility.

### DIFF
--- a/blosc/CMakeLists.txt
+++ b/blosc/CMakeLists.txt
@@ -3,6 +3,12 @@ add_definitions(-DUSING_CMAKE)
 
 set(INTERNAL_LIBS ${CMAKE_SOURCE_DIR}/internal-complibs)
 
+# Hide symbols by default unless they're specifically exported.
+# This makes it easier to keep the set of exported symbols the
+# same across all compilers/platforms.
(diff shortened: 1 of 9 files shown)


FrancescAlted:
I have had a look at the new visible symbols (basically, the ones listed in blosc.h) and I think they are all in.  So merging.

PR #124, opened 2015-05-10
AppVeyor CI configuration

jack-pappas:
I've added a configuration for AppVeyor CI. It's similar to travis-ci but geared towards Windows. It's free and runs alongside travis-ci, so now we'll be able to see whether commits/PRs will break Windows builds.

@FrancescAlted Like travis-ci, you'll need to sign in to AppVeyor via Github and add the c-blosc project to your account (again, it's free) so AppVeyor can set up the necessary integration with Github.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,13 @@
 # AppVeyor CI build configuration for c-blosc
 
+# Before cloning the repo, configure git to handle line endings correctly.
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Wow, we needed this very badly, thanks a lot.  Let's see how it goes.

PR #125, opened 2015-05-13
Retry of shuffle tests

jack-pappas:
This is a retry of #114. It turned out I'd forgotten to define a symbol (``BLOSC_TESTING``) in ``tests/CMakeLists.txt`` so even though the normally-hidden symbols were exported from the new "testing" build of blosc they weren't visible from the point of view of the test programs which caused compilation to fail.

I've tested this PR on Windows and Linux (Ubuntu 12.04 LTS amd64) and the tests now compile, link, and run correctly on both.


FrancescAlted:
Looks very good.  Thank you!

PR #127, opened 2015-05-17
Only default to a specific build type when using a single-configurati…

jack-pappas:
In the CMake build script, blosc currently defaults to the ``Release`` configuration if no build type has been explicitly specified (e.g., via ``-DCMAKE_BUILD_TYPE=Debug``). This behavior is fine for "single-configuration" generators like Makefiles; however, it causes some issues with multi-configuration generators like MSBuild / Visual Studio where the generated build scripts include all of the configurations (and the desired configuration is then chosen at build-time instead of configure-time).

This PR fixes the issue. I've tested on Ubuntu 12.04 LTS (same as travis-ci uses) and confirmed the behavior isn't changed there. On Windows + Visual Studio, the build type is no longer forced at configure-time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,15 +124,21 @@ endif(NOT DEACTIVATE_ZLIB)
 
 # create the config.h file
 configure_file ("blosc/config.h.in"  "blosc/config.h" )
+
 # now make sure that you set the build directory on your "Include" path when compiling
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Looks good to me.  Thanks for bringing your MSVC expertise!

PR #128, opened 2015-05-17
Split test_basics and make tests data-driven

jack-pappas:
The ``test_basics`` test program includes a fairly wide-ranging test suite now and is getting a bit unwieldy to work with. This PR addresses that issue by splitting ``test_basics`` into several smaller test programs aimed at testing specific pieces of the blosc code.

Some of the tests in ``test_basics`` include a number of test cases. When splitting these tests out into individual test programs, I also implemented functionality to make them data-driven. For each test program, ``tests/CMakeLists.txt`` now looks for an optional ``.csv`` file containing the parameters for each test case. This has a few benefits:

* It's now possible to see specifically which test cases failed just from the ctest log. Prior to this PR, if ``test_basics`` failed there was no way to determine which case failed without re-running the test program under a debugger or compiling in some printf statements. As an added benefit, if ``ctest`` is run with the ``-V`` option, the exact command-line used to run the test case is printed, making it trivial to reproduce an issue.
* The data-driven test programs don't need to be modified to add additional test cases. New test cases can be included by simply adding the parameters to the appropriate ``*.csv`` and re-running ``cmake`` (required because the ``*.csv`` file is read at configure-time).
* In ``test_basics``, if a single test case failed within any of the tests the program short-circuited to return the failure. This made it impossible to tell if any of the other test cases were also affected by some change being tested. Now that the individual test cases are separate, every test case is executed when the tests are run, regardless of whether other tests or test cases ran successfully.
* ``ctest`` includes functionality that makes it easy to select which test cases are run. For example, when developing a change that only impacts a certain part of blosc, you can now easily run just the tests on the affected code. (Of course, the whole suite should be run to verify the changes before committing them or submitting a PR).

The ``*.csv`` files for the tests currently contain the same test cases as what was hard-coded in ``test_basics`` before these changes, so test coverage should be exactly the same. Some parameters that were hard-coded to a single value before (e.g., compression level for the compression round-trip tests) are also included in the ``*.csv`` files so you can add additional test cases with different values for them.


FrancescAlted:
Pretty damn amazing.  Not sure if I like that much verbosity, but my intuition says that it is definitely the way to go.  Thanks!

PR #129, opened 2015-05-17
SSE2 shuffle support for irregular vector sizes

jack-pappas:
I've modified the SSE2 shuffle (and a minor modification to the generic shuffle) so if given a buffer containing a number of elements which isn't a multiple of the vector size, the vectorized version can still be used to process most of the elements before finishing up with the generic shuffle.

The performance should be the same for the cases where the vectorized shuffle was already being used. Performance of many common cases should be improved now since they'll be able to leverage vectorization no matter how many elements need to be shuffled.

### DIFF
--- a/blosc/shuffle-sse2.c
+++ b/blosc/shuffle-sse2.c
@@ -38,18 +38,18 @@ static void printxmm(__m128i xmm0)
 
 /* Routine optimized for shuffling a buffer for a type size of 2 bytes. */
 static void
-shuffle2_sse2(uint8_t* dest, const uint8_t* src, size_t size)
+shuffle2_sse2(uint8_t* const dest, const uint8_t* const src,
+  const size_t vectorizable_elements, const size_t total_elements)
 {
-  size_t i, j, k;
-  size_t numof16belem;
+  static const size_t bytesoftype = 2;
+  size_t j;
+  int k;
   __m128i xmm0[2], xmm1[2];
 
-  numof16belem = size / (16*2);
-  for (i = 0, j = 0; i < numof16belem; i++, j += 16*2) {
-    /* Fetch and transpose bytes, words and double words in groups of
-       32 bytes */
+  for (j = 0; j < vectorizable_elements; j += sizeof(__m128i)) {
+    /* Fetch 16 elements (32 bytes) then transpose bytes, words and double words. */
     for (k = 0; k < 2; k++) {
-      xmm0[k] = _mm_loadu_si128((__m128i*)(src+j+k*16));
+      xmm0[k] = _mm_loadu_si128((__m128i*)(src + (j * bytesoftype) + (k * sizeof(__m128i))));
       xmm0[k] = _mm_shufflelo_epi16(xmm0[k], 0xd8);
       xmm0[k] = _mm_shufflehi_epi16(xmm0[k], 0xd8);
       xmm0[k] = _mm_shuffle_epi32(xmm0[k], 0xd8);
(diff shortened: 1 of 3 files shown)


blosc/shuffle-sse2.c
```diff
@@ -66,63 +66,65 @@ shuffle2_sse2(uint8_t* dest, const uint8_t* src, size_t size)
...
+    for (j = 0; j < 2; j++) {
+      xmm0[j*2] = _mm_unpacklo_epi64(xmm1[j], xmm1[j+2]);
+      xmm0[j*2+1] = _mm_unpackhi_epi64(xmm1[j], xmm1[j+2]);
     }
     /* Store the result vectors */
-    for (k = 0; k < 4; k++) {
-      ((__m128i *)dest)[k*numof16belem+i] = xmm0[k];
+    uint8_t* const dest_for_ith_element = dest + i;
+    for (j = 0; j < 4; j++) {
+      _mm_storeu_si128((__m128i*)(dest_for_ith_element + (j * total_elements)), xmm0[j]);
```
FrancescAlted: Why you replaced aligned stores by unaligned ones?  That is making my benchmarks a bit slower (5%) and, while not too bad, I am curious on the change.

FrancescAlted: Well, I just spent a bit more time on this and redid my benchmarks and I am not seeing the slow-down anymore, or a very small difference (< 1%) (using a i5-3380M @ 2.90GHz).

blosc/shuffle-sse2.c
```diff
@@ -409,35 +439,56 @@ shuffle_sse2(const size_t bytesoftype, const size_t blocksize,
...
-  if (unaligned_src || unaligned_dest || !multiple_of_block || too_small) {
-    /* _src or _dest buffer is not aligned, not multiple of the vectorization
-     * size or is not too small.  Call the generic routine. */
+  const size_t vectorized_chunk_size = bytesoftype * sizeof(__m128i);
+  
+  int unaligned_src = (int)((uintptr_t)_src % sizeof(__m128i));
+  int unaligned_dest = (int)((uintptr_t)_dest % sizeof(__m128i));
+  int too_small = (blocksize < vectorized_chunk_size);
+
+  if (unaligned_src || unaligned_dest || too_small) {
```
FrancescAlted: Provided that all the stores are made through `storeu`, maybe we can remove the 'unaligned_dest' condition here?

jack-pappas:
@FrancescAlted I'm happy to make the change either way you want. I can change the loads/stores back to aligned to preserve the current behavior for now (so we can investigate it separately), or I can just remove the alignment check for the buffers. Just let me know what you think is best and I'll make the change.

FrancescAlted:
I would prefer to separate concerns and let the alignment modifications out for now.  Later we can go back to alignment and, with some benchmarks in hand, make a more informed decisions about this.

jack-pappas:
I tried replacing the ``_mm_loadu_si128`` calls with ``_mm_load_si128`` and the ``_mm_storeu_si128`` calls with ``_mm_store_si128``. After doing so, I compiled blosc and ran the tests -- and some of the *test_shuffle_roundtrip_sse2* tests segfaulted. I confirmed the segfaults occur with both Windows+MSVC and Linux+GCC.

``jack@jack-linux64:~/SourceCode/c-blosc/build$ ctest -R test_shuffle_roundtrip_sse2``

```
94% tests passed, 24 tests failed out of 399

Total Test time (real) =  16.69 sec

The following tests FAILED:
	 33 - test_shuffle_roundtrip_sse2_2_500_32_0 (SEGFAULT)
	 34 - test_shuffle_roundtrip_sse2_2_500_32_1 (SEGFAULT)
	 35 - test_shuffle_roundtrip_sse2_2_500_32_2 (SEGFAULT)
	 42 - test_shuffle_roundtrip_sse2_2_702713_32_0 (SEGFAULT)
	 43 - test_shuffle_roundtrip_sse2_2_702713_32_1 (SEGFAULT)
	 44 - test_shuffle_roundtrip_sse2_2_702713_32_2 (SEGFAULT)
	 75 - test_shuffle_roundtrip_sse2_4_500_32_0 (SEGFAULT)
	 76 - test_shuffle_roundtrip_sse2_4_500_32_1 (SEGFAULT)
	 77 - test_shuffle_roundtrip_sse2_4_500_32_2 (SEGFAULT)
	 84 - test_shuffle_roundtrip_sse2_4_702713_32_0 (SEGFAULT)
	 85 - test_shuffle_roundtrip_sse2_4_702713_32_1 (SEGFAULT)
	 86 - test_shuffle_roundtrip_sse2_4_702713_32_2 (SEGFAULT)
	159 - test_shuffle_roundtrip_sse2_8_500_32_0 (SEGFAULT)
	160 - test_shuffle_roundtrip_sse2_8_500_32_1 (SEGFAULT)
	161 - test_shuffle_roundtrip_sse2_8_500_32_2 (SEGFAULT)
	168 - test_shuffle_roundtrip_sse2_8_702713_32_0 (SEGFAULT)
	169 - test_shuffle_roundtrip_sse2_8_702713_32_1 (SEGFAULT)
	170 - test_shuffle_roundtrip_sse2_8_702713_32_2 (SEGFAULT)
	201 - test_shuffle_roundtrip_sse2_16_500_32_0 (SEGFAULT)
	202 - test_shuffle_roundtrip_sse2_16_500_32_1 (SEGFAULT)
	203 - test_shuffle_roundtrip_sse2_16_500_32_2 (SEGFAULT)
	210 - test_shuffle_roundtrip_sse2_16_702713_32_0 (SEGFAULT)
	211 - test_shuffle_roundtrip_sse2_16_702713_32_1 (SEGFAULT)
	212 - test_shuffle_roundtrip_sse2_16_702713_32_2 (SEGFAULT)
```

Digging deeper, I found I'm able to change the loads in the shuffle functions to aligned loads and the stores in the unshuffle functions to aligned stores without triggering the segfaults. It's only when the shuffle functions use aligned stores and/or the unshuffle functions use aligned loads that those tests segfault.

I compiled blosc from the ``master`` branch then disassembled it like so:

``jack@jack-linux64:~/SourceCode/c-blosc/build$ objdump -S -M intel-mnemonic blosc/CMakeFiles/blosc_shared_testing.dir/shuffle-sse2.c.o``

The disassembly shows that GCC generates a mix of ``movdqa`` and ``movdqu`` instructions within the SSE2 shuffle/unshuffle functions. When I compile the code from this PR, I see only ``movdqu`` instructions (no ``movdqa``). If I make the changes to use only the aligned load/store intrinsics, compile, and disassemble, I only see ``movdqa`` instructions (however, this is the version that causes the tests to segfault).

In other words, it looks like blosc is already using *some* unaligned loads/stores and won't work without them. That being the case, how do you want to proceed? I think the performance impact from switching to all unaligned loads/stores will be small (if even noticeable), especially when compared against the performance benefits from the changes in this PR and the follow-up PR I have in store.

FrancescAlted:
Ok, in that case it would better to revisit #105 and do some benchmarks and see if we go with unalignment for everything.  I have a box with a Core2 and another with a modern i5, so it will be nice to see the differences in both platforms.  If others want to contribute to that benchmarking you are welcome!

FrancescAlted:
I have decided to use unaligned access for SSE2 operation (AVX2 was using that already), so feel free to rebase and let's merge this afterwards.  Thanks!

jack-pappas:
@FrancescAlted I've rebased the changes onto master (after your merge of #130) and squashed them into a single commit, so this can be merged whenever you're ready.

FrancescAlted:
Looks good to me, thanks!  I just realized that the AVX2 code would need something similar.  Would you like to port the job here to AVX2 too?

jack-pappas:
Yes, I'm planning to port the changes made to the SSE2 implementation in this PR and #133 to the AVX2 implementation as well.

FrancescAlted:
Sounds very good.

PR #130, opened 2015-05-25
Updated version of unaligned load and store with SSE2. Overrides PR #105

FrancescAlted:


FrancescAlted:
I have benchmarked a laptop with an i5-3380M @ 2.90GHz, and I have not found differences in performance due to the use of unaligned loads ('loadu') and stores ('storeu').

FrancescAlted:
With another laptop (Core2 @ 2.13 MHz), I can notice slowdowns due to unalignment stores and loads up to a 30% for compressing and up to 20% for decompressing.  This is compatible with what @juliantaylor is reporting.

FrancescAlted:
With a Xeon E3-1240 v3 @ 3.40GHz I cannot see any difference in performance either.

juliantaylor:
anything newer than sandy bridge has no penalty, anything older has, don't know where the cutoff for amd's is though. A phenom2 has penalty.

FrancescAlted:
Yes, Sandy Bridge seems to be the cutoff for Intel.

FrancescAlted:
The thing is that I should make decision on merging this because @jack-pappas is waiting for this before being able to apply #129 (and others that may come later).

FrancescAlted:
Well, provided that this patch allows to apply SSE2 acceleration even for 8-bytes aligned buffers in 32-bit machines, this can somewhat compensate the slowdown on old CPUs, besides allowing @jack-pappas making more progress, so hereby I am taking the decision to merge this in.

juliantaylor:
I am a bit concerned about amd cpus I'm not sure if there even is one without penalty.
I don't know a significant use case where this gains anything. On legacy 32 bit arches sse2 is disabled anyway.

FrancescAlted:
As said, allowing Jack making progress in #129 is important.  And apparently he needs unaligned support for that.  Making c-blosc slower in older CPUs is an unfortunate side-effect, but I think the benefits are larger.

juliantaylor:
peeling to alignment is not an option?

FrancescAlted:
Judging by Jack's comment: https://github.com/Blosc/c-blosc/pull/129#issuecomment-104083823, apparently mixing unalignment and alignment is giving problems.  I am afraid that looking into this is out of my expertise, but are very welcome if you want to dig deeper into that one.

FrancescAlted:
Ok, I don't want to keep this waiting for too long, so I am merging this so that @jack-pappas can make progress on #129 .  I really think that the new scenarios where SSE2 can be used will more than compensate the slowness on unaligned access in old CPUs.

PR #133, opened 2015-05-27
Added SSE2 shuffle/unshuffle routines for types larger than 16 bytes

jack-pappas:
This PR is a re-submission of #88, part of which was already included in #129. This adds two additional shuffle/unshuffle functions to the SSE2 implementation which extend the vectorization support to types larger than 16 bytes.


FrancescAlted:
Merging.

PR #135, opened 2015-05-29
Implemented peeled AVX2 shuffle

jack-pappas:
Modified the AVX2 shuffle implementation to use the same approach as the "peeled" SSE2 implementation submitted in #129. This allows the AVX2 shuffle to be used even when the number of elements in the buffer is not a multiple of the vector size (in this case, 32).

This commit also fixes an addressing bug in ``shuffle2_avx2`` which caused it to store the shuffled data incorrectly for certain input buffer sizes.

Note: this PR should be tested manually on an AVX2-capable system by uncommenting the few lines in ``tests/CMakeLists.txt`` that enable the AVX2 tests. I've done this locally and the tests pass, but someone else should confirm on their system before merging this since neither of the CI builds currently support AVX2 (so the tests that do run won't cover these changes).

### DIFF
--- a/blosc/shuffle-avx2.c
+++ b/blosc/shuffle-avx2.c
@@ -46,18 +46,19 @@ shuffle2_avx2(uint8_t* const dest, const uint8_t* const src,
   const size_t vectorizable_elements, const size_t total_elements)
 {
   static const size_t bytesoftype = 2;
-  static const uint8_t b_mask[] = {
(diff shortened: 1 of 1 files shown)


FrancescAlted:
This is very nice Jack.  I have tested that on a AVX2 machine under Linux (uncommenting the lines in `tests/CMakeLists.txt`), and all tests pass, so merging.  Thanks a lot!

PR #136, opened 2015-05-30
AVX2 shuffle support for types larger than 16 bytes

jack-pappas:
This PR adds support to the AVX2 shuffle for types larger than 16 bytes. It's basically an AVX2 version of the code added to the SSE2 shuffle in #133.

As in #135, this PR should be tested manually on an AVX2-capable system by uncommenting the few lines in tests/CMakeLists.txt that enable the AVX2 tests. I've done this locally and the tests pass, but someone else should confirm on their system before merging this since neither of the CI builds currently support AVX2 (so the tests that do run won't cover these changes).

### DIFF
--- a/blosc/shuffle-avx2.c
+++ b/blosc/shuffle-avx2.c
@@ -39,6 +39,27 @@ static void printymm(__m256i ymm0)
 }
 #endif
 
+/* GCC and older versions of Clang don't include the split load/store
+   intrinsics needed for the tiled shuffle, so defined them here if necessary. */
+#if !defined(_mm256_loadu2_m128i)
+static inline __m256i
+__attribute__((__always_inline__, __nodebug__))
+_mm256_loadu2_m128i(const __m128i* const hiaddr, const __m128i* const loaddr)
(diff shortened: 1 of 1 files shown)


jack-pappas:
I've defined those missing functions for GCC, so the build error should be fixed now. Would you mind giving it another try?

FrancescAlted:
I cannot do that today, but this morning I tried your solution and there was another issue popping out (don't remember the details sorry).  I think the best would be for you to have access to a AVX2 machine and use a GCC (or MinGW).  If you don't have access to a box with physical AVX2, maybe QEMU would be a workaround?

jack-pappas:
My development machine is a Haswell, so I've been testing with AVX2 already. I've tested these changes with both VS2013 and GCC 4.9.2 (via MinGW-w64) on Windows 7 x64. I just figured you'd want to confirm. I think the changes are fine to merge whenever you have a chance to do so.

FrancescAlted:
Ok, I will merge then.  Thanks!

PR #141, opened 2015-06-17
Update blosc.c

atchouprakov:
Uninitialized member context.threads_started, causes crash (tries to destroy nonexistent thread pool)

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -1248,6 +1248,7 @@ int blosc_decompress_ctx(const void *src, void *dest, size_t destsize,
 			 int numinternalthreads)
 {
   struct blosc_context context;
+  context.threads_started = 0;
   int result = blosc_run_decompression_with_context(&context, src, dest, destsize, numinternalthreads);
 
   if (numinternalthreads > 1)


FrancescAlted:
Looks good to me.  Could you please provide a minimal example showing the crash so that we can add it to the test suite?  Thanks!

FrancescAlted:
I am going to merge this now.  Please feel free to send the simple crashing example later (if you can).

atchouprakov:
I don't know how to create a test for it. To make a crash you have to force  context.threads_started to be > 0 before call to blosc_run_decompression_with_context , for example if I modify blosc_decompress_ctx to:
```c
/* The public routine for decompression with context. */
int blosc_decompress_ctx(const void *src, void *dest, size_t destsize,	 int numinternalthreads)
{
  struct blosc_context context;
  //context.threads_started = 0;
  context.threads_started = 42;
  int result = blosc_run_decompression_with_context(&context, src, dest, destsize, numinternalthreads);
....
}
```
I have 100% crash on MS VS2013 x64 with numinternalthreads > 1.
but blosc_context and blosc_run_decompression_with_context are private and not currently accessible in the test framework.
BTW it might be a good idea to expose context creation/destruction to public API, so you can create one context per application thread and reuse it versus recreating it every time. I can implement it, if it is of interest to others.

PR #144, opened 2015-06-19
travis_badge

FrancescElies:
Travis badge on the front page


FrancescElies:
You could check it here in case you would like to https://github.com/FrancescElies/c-blosc/tree/travis_badge

FrancescAlted:
Looks good.

PR #153, opened 2015-09-07
[MRG] Make set_nthreads a noop

sdvillal:
If the number of threads is not changing, do not reinitialise the thread pool in ```blosc_set_nthreads```. I think this should be a harmless change, but I'm not familiar enough with the code to be sure it is a good idea.

Context: I'm compressing many small pieces of independent data. For each of these compression calls, I set the number of threads to a configured value - to avoid depending on uncontrolled global state. Surprisingly, when the number of calls is large (let's say 400K) and the compressed data is not so large (let's say, 16K per call), the time wasted on pool reinitialisation becomes noticeable.

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -1616,11 +1616,12 @@ int blosc_set_nthreads(int nthreads_new)
 {
   int ret = g_threads;
 
-  /* Re-initialize Blosc */
-  blosc_destroy();
-  blosc_init();
-
-  g_threads = nthreads_new;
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Yes, the change looks good to me (although I *vaguely* remember having a similar code before but had to remove that, but I am uncertain about my memories :(

I am +1 for this, but would like to know the kind of speed-up that we are talking about.  Do you have some figures?

FrancescAlted:
Merging this as it seems harmful.

PR #160, opened 2016-03-06
Fix double destroy of blosc context

dswanepoel:
The global context is destroyed whenever blosc_set_nthreads() is called, so explicitly destroying it in the loop causes blosc_set_nthreads() to try to destroy it again, leading to a glibc double free error. Moving blosc_destroy() out of the loop prevents the double free.


FrancescAlted:
Sure thing.  Thanks.

PR #161, opened 2016-03-22
Implement _xgetbv() intrinsic for VS2008 compatibility

jack-pappas:
I saw an email from @FrancescAlted regarding c-blosc and bcolz compatibility issues related to recent versions of c-blosc no longer compiling with VS2008 (due to the use of the ``_xgetbv()`` intrinsic, which it doesn't support).

I had a go at fixing the shuffle code to implement ``_xgetbv()`` when compiling with VS2008. I haven't tested this code at all (I don't have vs2008 tools installed), so it may not even compile, or it may compile but fail at run-time. If it works though, it should fix the compatibility issues with bcolz (and other Python 2.x libraries).

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -89,8 +89,49 @@ static blosc_cpu_features blosc_get_cpu_features(void) {
 #else
 
 #if defined(_MSC_VER) && !defined(__clang__)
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Great!  Merging to see how the new VS2008/VS2010 support in appveyor config file goes.

PR #166, opened 2016-05-03
Allow hardware features to be specified over the command-line

jbrd:
In my particular case, I would like to be able to prepare a build of blosc on new hardware (an AVX2 compatible machine), but have this run on older hardware (e.g. machines without AVX2 support).

The existing CMakeLists.txt file suggests that some choices were made to permit cross-compilation, however I couldn't find any existing options to allow me to do this without patching the actual CMakeLists.txt file.

The change I have made here is to ensure that the existing auto-detection only runs if both `COMPILER_SUPPORT_SSE2` and `COMPILER_SUPPORT_AVX2` are not already defined. To make things easier for the user, I have also covered the case where only one of these is specified over the command-line (in which case, a sensible default is chosen for the missing variable).

I should note that in my case, I am integrating blosc into our build system here, which prefers the passing of command-line options over patching the source tree.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,7 +159,17 @@ endif()
 # by this (the host) system, because we want to be able to support compiling
 # for newer hardware on older machines as well as cross-compilation.
 message(STATUS "Building for system processor ${CMAKE_SYSTEM_PROCESSOR}")
-if(CMAKE_SYSTEM_PROCESSOR STREQUAL i386 OR
+
+# Allow hardware features to be specified over the command-line (e.g. -DCOMPILER_SUPPORT_AVX2).
+if(DEFINED COMPILER_SUPPORT_SSE2 AND NOT DEFINED COMPILER_SUPPORT_AVX2)
+    set(COMPILER_SUPPORT_AVX2 FALSE)
+elseif(DEFINED COMPILER_SUPPORT_AVX2 AND NOT DEFINED COMPILER_SUPPORT_SSE2)
+    set(COMPILER_SUPPORT_SSE2 ${COMPILER_SUPPORT_AVX2})
+endif()
+
+if(DEFINED COMPILER_SUPPORT_SSE2 AND DEFINED COMPILER_SUPPORT_AVX2)
+    message(STATUS "Taking hardware features from command-line (SSE2=${COMPILER_SUPPORT_SSE2}, AVX2=${COMPILER_SUPPORT_AVX2}).")
+elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL i386 OR
    CMAKE_SYSTEM_PROCESSOR STREQUAL i686 OR
    CMAKE_SYSTEM_PROCESSOR STREQUAL x86_64 OR
    CMAKE_SYSTEM_PROCESSOR STREQUAL amd64 OR


FrancescAlted:
C-Blosc already has [*run-time* detection of SSE2 and AVX2](https://github.com/Blosc/c-blosc/blob/master/blosc/shuffle.c#L300), so even if your compiler adds AVX2 support the code should be able to run on a non-AVX2 processor.  Is that not working for you?

jbrd:
Thanks for such a fast response, and apologies for my late one! I did run into some problems out-the-box, but these are actually build-time problems rather than run-time problems (so you're right, I was incorrect in my assumption that my AVX machine was the cause of these issues).

We're building in a software environment with gcc 4.9.1 and so by default, AVX2 support is enabled. However, on machines without AVX2 support, we're getting the following errors at build-time:

`[ 12%] Building C object blosc/CMakeFiles/blosc_shared.dir/shuffle-avx2.c.o
/user_data/.tmp/bobscratch-7FQ7ig/tmp/cc6V1wbZ.s: Assembler messages:
/user_data/.tmp/ccm1nldE.s:62: Error: suffix or operands invalid for `vpunpckhqdq'
/user_data/.tmp/ccm1nldE.s:63: Error: no such instruction: `vpermq $216,%ymm11,%ymm7'
`

The following command is being used to build this file:

`
cd /hosts/dist/dist/blosc/c-blosc-1.9.0/blosc && /builds/gcc/4.9.1/d9048547a5/bin/cc  -DBLOSC_SHARED_LIBRARY -DUSING_CMAKE -Dblosc_shared_EXPORTS -msse2 -O3 -DNDEBUG -fPIC -fvisibility=hidden -I/hosts/dist/dist/blosc/c-blosc-1.9.0/blosc -I/hosts/dist/dist/blosc/c-blosc-1.9.0/internal-complibs/lz4-1.7.2 -I/hosts/dist/dist/blosc/c-blosc-1.9.0/internal-complibs/snappy-1.1.1    -mavx2 -o CMakeFiles/blosc_shared.dir/shuffle-avx2.c.o   -c /hosts/dist/dist/blosc/c-blosc-1.9.0/blosc/shuffle-avx2.c
`

So the changes in this commit allows us to effectively work around this problem by explicitly disabling AVX2 support at build-time, which has been useful for us. Perhaps there is a better solution for this problem though?

FrancescAlted:
Hmm, this is a bug indeed.  The good solution is to have real SSE2 / AVX2 detection working, but apparently it is not.  @juliantaylor what do you think about this?  It could be fixed easily?  If not, I can merge this PR as a temporary workaround.

FrancescAlted:
Pinging @jack-pappas too.

juliantaylor:
is this really gcc 4.9? that support avx2, it was added around 4.7 or 4.8

I tested it with gcc 4.4 which does not support it and it correctly skips building the avx2 files.

juliantaylor:
that said the configure check is just a simple version range check which could be replaced with a more robust test compile check if necessary

FrancescAlted:
Thanks @juliantaylor. And sorry, I misread the original posts and I thought the problem was at runtime, not compile time.  @jbrd I regularly compile C-Blosc on a non-AVX2 box (Travis CI does not have that either) and I have no problems.  I used gcc 4.9.1 previously without issues too (now switched to 5.3.1).  Could you double check you compiler versions in the non-AVX2 box and report back?

juliantaylor:
from the error message it looks like a custom gcc installation. I assume that the compiler itself is 4.9 but it somehow uses a different assembler, possibly the older system one. So also check which assembler it ends up using (typically the binary is named `as`, check what is used e.g. via strace -f -e execve gcc somefile.c -c)

jbrd:
Thanks for the suggestions everyone, I will go back and take a look at our build of gcc - it is indeed built from source, so I'll try to figure out exactly why this is happening and post back when I have more information...

juliantaylor:
I think you need binutils 2.22 for avx2 support, RHEL6 version is too old so maybe this is something one should check better in the configuration.
for RHEL6 you can use their developer toolsets for a newer gcc and assembler.

jbrd:
@juliantaylor you are absolutely right, we are locked down to an older version of binutils which doesn't have AVX2 support. Unfortunately upgrading binutils across our facility is going to be quite involved I think, so in the short term, I think I'm going to have to continue to apply this patch internally. Happy to reject this PR - it will simply require more work on our side when we ingest new versions of blosc. Factoring the binutils version requirement into the existing auto-detection mechanism would make our lives a bit easier, I'd be happy to contribute this if you feel its worth the effort?

FrancescAlted:
What about providing a PR where you explicitly can *disable* AVX2 via something like '-DNO_AVX2_SUPPORT'?  This is less intrusive and you can always assert it for your toolchain.

jbrd:
Yes that sounds like a good plan, will push up another PR shortly...

PR #167, opened 2016-05-04
Fix errors when building on multiple cores (e.g. -jN in gcc)

jbrd:
Running make with the -jN option was causing the tests to be built before the blosc_testing library had been built, resulting in linker errors. Adding this dependency fixes the issue...


FrancescAlted:
Great.  I have been wondering for some time why the `make -j N` did not work.  Thanks!

PR #168, opened 2016-05-04
Added DEACTIVATE_AVX2 cmake option...

jbrd:
…to explicitly disable AVX2 at build-time


FrancescAlted:
Looks good to me.  Thanks!

PR #173, opened 2016-07-28
Compilation warnings

keszybz:


FrancescAlted:
Hmm, I think you need to cope with the Visual Studio way to deal with `inline`.  However, it would be nice if you do some benchmark prior to send this kind of optimization PRs; sometimes you can get interesting surprises ;)

OTOH, I have removed the unused variables in a8db55b.

PR #176, opened 2016-09-04
preserve C_FLAGS from environment

danstender:
Little patch for preserving CMAKE_C_FLAGS from the environment while setting std=gnu99.
Best,
DS


FrancescAlted:
LGTM

PR #178, opened 2016-10-07
CMake tweaks to allow subproject use

kdmurray91:
Hi,

This PR makes several minor tweaks to the CMake configs to enable projects to more easily include blosc as CMake subprojects (included in-tree with git submodules or subtree, or just `cp`).

Notably:

- Include `BUILD_SHARED` option which corresponds to `BUILD_STATIC`, enabling the shared library build. This defaults to ON
- Use `${PROJECT_SOURCE_DIR}` where appropriate
- Add a `blosc` target alias which lets people simply do `target_link_libraries(my_thing thing_lib blosc)` to get blosc compiled in as a static library
- Add `BLOSC_INSTALL` option to control installation. Defaults to on.
- Only set up CPack if blosc is NOT being used as a sub-project.

I have confirmed that both normal installation and use as a sub-project works after these changes. I can prepare a demonstration of how I used this in a project of mine if you'd like.

Cheers,
K

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 bench/bench
 build/
+
+*.sw?
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -297,11 +297,6 @@ endif()
 # subdirectories
 add_subdirectory(blosc)
 
-if (BLOSC_IS_SUBPROJECT)
-    add_library(blosc ALIAS blosc_static)
-    message(STATUS "Enabled `blosc` alias target to static library.")
-endif()
-
 if(BUILD_TESTS)
     enable_testing()
     add_subdirectory(tests)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,8 @@
 #
 #   BUILD_STATIC: default ON
 #       build the static version of the Blosc library
+#   BUILD_SHARED: default ON
+#       build the shared library version of the Blosc library
 #   BUILD_TESTS: default ON
 #       build test programs and generates the "test" target
 #   BUILD_BENCHMARKS: default ON
@@ -84,6 +86,8 @@ message("Configuring for Blosc version: " ${BLOSC_VERSION_STRING})
 # options
 option(BUILD_STATIC
     "Build a static version of the blosc library." ON)
+option(BUILD_SHARED
+    "Build a shared library version of the blosc library." ON)
 option(BUILD_TESTS
     "Build test programs form the blosc compression library" ON)
 option(BUILD_BENCHMARKS
@@ -107,7 +111,7 @@ option(PREFER_EXTERNAL_ZLIB
 option(PREFER_EXTERNAL_ZSTD
     "Find and use external Zstd library instead of included sources." ON)
 
-set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 
 
 if(NOT DEACTIVATE_LZ4)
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,2 +1,4 @@
 bench/bench
 build/
+
+*.sw?
```
kdmurray91: Ah, whoops. This wasn't meant to be committed, but hey, my fellow Vim users probably won't mind :). I can remove it if you want.

blosc/CMakeLists.txt
```diff
@@ -1,12 +1,12 @@
...
 add_definitions(-DUSING_CMAKE)
 
-set(INTERNAL_LIBS ${CMAKE_SOURCE_DIR}/internal-complibs)
+set(INTERNAL_LIBS ${PROJECT_SOURCE_DIR}/internal-complibs)
 
 # Hide symbols by default unless they're specifically exported.
 # This makes it easier to keep the set of exported symbols the
 # same across all compilers/platforms.
-set(CMAKE_C_VISIBILITY_PRESET hidden)
+#set(CMAKE_C_VISIBILITY_PRESET hidden)
```
FrancescAlted: Why is CMAKE_C_VISIBILITY_PRESET = hidden is gone?

.gitignore
FrancescAlted: Nah, this is fine.

CMakeLists.txt
```diff
@@ -272,10 +276,32 @@ if(WIN32)
...
+    endif()
+endif()
+
 
 # subdirectories
 add_subdirectory(blosc)
 
+if (BLOSC_IS_SUBPROJECT)
+    add_library(blosc ALIAS blosc_static)
+    message(STATUS "Enabled `blosc` alias target to static library.")
```
FrancescAlted: Why blosc needs in this case to be an alias of the static library and not of the shared one?  Just curious.

blosc/CMakeLists.txt
kdmurray91: Ah, another stuff-up on my part. According to a CMake warning, this should only be enabled for the shared lib, and I was going to work out how to set this in a target specific way, but obviously forgot. I can either re-enable it if that warning is irrelvant, or work out how to make it SO-specific

CMakeLists.txt
kdmurray91: Ah, this follows a pattern from `gtest`. I actually use `blosc_static` anyway, for extra peace of mind, so I'm happy to remove if you find this unnecessary.

kdmurray91:
@FrancescAlted Thanks!

I plan to use this as follows. I have a git subtree of the c-blosc repo under `./ext/blosc`.

In **my** root-level `CMakeLists.txt`:
```
SET(DEACTIVATE_SNAPPY ON CACHE BOOL "")
SET(DEACTIVATE_ZSTD ON CACHE BOOL "")
SET(DEACTIVATE_ZLIB ON CACHE BOOL "")
SET(BUILD_SHARED OFF CACHE BOOL "")
SET(BUILD_TESTS OFF CACHE BOOL "")
SET(BUILD_BENCHMARKS OFF CACHE BOOL "")
ADD_SUBDIRECTORY(ext/blosc)
```

In `src/CMakeLists.txt`:

```
ADD_EXECUTABLE(mythingy ${THINGY_SRCS})
TARGET_LINK_LIBRARIES(mythingy ${DEP_LIBS} blosc_static)
```

I also use the HDF5 filter, but for now this just gets added to `${THINGY_SRCS}`.

blosc/CMakeLists.txt
FrancescAlted: I haven't seen this warning, but if so, I think we should enable it just for the shared lib then.

CMakeLists.txt
FrancescAlted: Yeah, removing this would be fine.

blosc/CMakeLists.txt
kdmurray91: For now, I've enabled it for all, as I'm about to leave work. I will try to work this out next week, possibly in a different PR.

CMakeLists.txt
kdmurray91: Done

FrancescAlted:
Thanks for the use case.  Looking at this, it would have been nice to use e.g. 'BLOSC_DEACTIVATE_SNAPPY' instead of 'DEACTIVATE_SNAPPY'.  But that would possible break some projects out there, so, perhaps just a good idea for the Blosc2 project.

FrancescAlted:
Also, would you mind updating the 'RELEASE_NOTES.txt' for your additions?

kdmurray91:
Certainly would. Likewise for BUILD_SHARED/STATIC

FrancescAlted:
@kdmurray91 Thanks for addressing my concerns!

PR #180, opened 2016-10-08
Add a pkg-config file to installed prefix

kdmurray91:
Hi,

This PR simply adds a pkg-config definition file, such that programs using pkg-config can find blosc and set their CFLAGS and LDFLAGS appropriately.

Cheers,
Kevin


FrancescAlted:
LGTM

PR #183, opened 2016-11-15
fix build on non-linux archs

danstender:
This patch is from Debian packaging to fix build breaks on non-Linux archs (thehurd, kfreebsd), please see <https://bugs.debian.org/844047>. It's confirmed to work for hurd-i386 (kfreebsd-s are pending). If you would like to accept that patch, welcome.
Best,
DS

### DIFF
--- a/bench/bench.c
+++ b/bench/bench.c
@@ -31,14 +31,14 @@
 #if defined(_WIN32)
   /* For QueryPerformanceCounter(), etc. */
   #include <windows.h>
-#elif defined(__MACH__)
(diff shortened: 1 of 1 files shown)


danstender:
... with that, builds also on the kfreebsd-s (<https://buildd.debian.org/status/package.php?p=c-blosc&suite=unstable>).

FrancescAlted:
LGTM.  Thanks!

FrancescAlted:
Oops, that was closed by mistake.  This PR has been committed in 254ffdf.

PR #188, opened 2017-02-27
Fix blosc_test_malloc on FreeBSD

iblis17:
The c compiler in FreeBSD system base is clang.
Seems that it do not have the macro `_ISOC11_SOURCE` but `__STDC_VERSION__`.

### DIFF
--- a/tests/test_common.h
+++ b/tests/test_common.h
@@ -62,7 +62,7 @@ static void* blosc_test_malloc(const size_t alignment, const size_t size)
   void *block = NULL;
   int32_t res = 0;
 
-#if _ISOC11_SOURCE
+#if _ISOC11_SOURCE || (__STDC_VERSION__ >= 201112L && !defined(__APPLE__))
   /* C11 aligned allocation. 'size' must be a multiple of the alignment. */
   block = aligned_alloc(alignment, size);
 #elif defined(_WIN32)


FrancescAlted:
Ok.  Thanks for the fix.

tests/test_common.h
```diff
@@ -62,7 +62,7 @@ static void* blosc_test_malloc(const size_t alignment, const size_t size)
   void *block = NULL;
   int32_t res = 0;
 
-#if _ISOC11_SOURCE
+#if _ISOC11_SOURCE || (__STDC_VERSION__ >= 201112L && !defined(__APPLE__))
```
FrancescAlted: @iblis17 Recently someone complained about this patch preventing to compile blosc on CentOS.  Could you tell me if this other patch would work on FreeBSD?

```
-#if _ISOC11_SOURCE
+#if defined(_ISOC11_SOURCE)
```

Thanks in advance.

iblis17: well, it doesn't work.

here are the predefined macros on my machine
```
clang -O2 -pipe -std=c11 -dM -E - < /dev/null
```
https://gist.github.com/iblis17/5881e22b304f00fd90cd446b5f007449

There isn't any `_ISO*` stuffs.
Or... can we use the macro `__FreeBSD__`?

FrancescAlted: Good idea.  Could you try with the `__FreeBSD__` macro?

iblis17: This works fine for me, and the tests passed on master
```diff
diff --git a/tests/test_common.h b/tests/test_common.h
index 4e40533..ce1be12 100644
--- a/tests/test_common.h
+++ b/tests/test_common.h
@@ -62,7 +62,7 @@ static void* blosc_test_malloc(const size_t alignment, const size_t size)
   void *block = NULL;
   int32_t res = 0;

-#if _ISOC11_SOURCE || (__STDC_VERSION__ >= 201112L && !defined(__APPLE__))
+#if defined(_ISOC11_SOURCE) || (defined(__FreeBSD__) && __STDC_VERSION__ >= 201112L)
   /* C11 aligned allocation. 'size' must be a multiple of the alignment. */
   block = aligned_alloc(alignment, size);
 #elif defined(_WIN32)
```

FrancescAlted: Great.  Thanks!

FrancescAlted: Could you please send a new PR for this?  Thanks in advance.

PR #204, opened 2017-10-25
prevent installing shared libraries if there are not built

adriankrupa:
Building only static libraries causes an error.


FrancescAlted:
Cool.  Thanks!

PR #205, opened 2017-12-30
Blosclz opt

FrancescAlted:
Optimization of the blosclz codec, mainly via using the memory copy functions in zlib-ng.  The result is a quite noticeable speedup, independently if using GCC, Clang or MSVC.


FrancescAlted:
I am done we that.  The performance in both speed and compression ratio has improved quite a bit.  This deserves a blog when I'd find some time.

PR #211, opened 2018-02-01
Fix benchmarks link in README

PeteHaitch:
This page seems to have moved, getting a 404 with old link


FrancescAlted:
Sure.  Thanks!

PR #213, opened 2018-02-02
Stable channel by default

lasote:
For the CI, when you push the tag, the channel was overwritten to "testing", just the opposite we pretended. With this PR all the packages will have the **stable** channel in CI (only uploaded when you tag).


FrancescAlted:
Ok, let's try that.

Issue #215, opened 2018-02-14
Data compressed with compressor other than BloscLZ using 1.11.1 not decompressible with 1.7.0


estan:
In migrating from Ubuntu 16.04 to Ubuntu 17.10, we've run into a snag with libblosc.

It seems that data compressed with libblosc 1.11.1 (the version in Ubuntu 17.10) is not decompressible using libblosc 1.7.0 (the version in Ubuntu 16.04) if using a compressor != BloscLZ.

Below is a test program (modified version of `examples/simple.c`).

To reproduce the problem (using the LZ4 compressor here as an example):

1. On Ubuntu 17.10 (libblosc 1.11.1), create a compressed file `test.blosc`:

```
estan@balder-101:~$ ./simple compress lz4 test.blosc
Blosc version info: 1.11.1 ($Date:: 2016-09-03 #$)
Compression: 4000000 -> 142070 (28.2x)
Wrote test.blosc
```

2. Copy the file to an Ubuntu 16.04 (libblosc 1.7.0) machine and try to decompress it:

```
[estan@newton ~]$ ./simple decompress lz4 test.blosc 
Blosc version info: 1.7.0 ($Date:: 2015-07-05 #$)
Read test.blosc
Decompression error.  Error code: -1
[estan@newton ~]$
```

The same can be seen when trying the `zlib` compressor. However (!): Using the `"blosclz"` compressor when compressing the file, the decompression succeeds.

**Info from the Ubuntu 16.04 system**

```
[estan@newton ~]$ dpkg -l | egrep "libblosc|liblz4"
ii  libblosc-dev                                    1.7.0-1                                                  amd64        High performance meta-compressor optimized for binary data (development files)
ii  libblosc1                                       1.7.0-1                                                  amd64        High performance meta-compressor optimized for binary data
ii  liblz4-1:amd64                                  0.0~r131-2ubuntu2                                        amd64        Fast LZ compression algorithm library - runtime
ii  liblz4-dev:amd64                                0.0~r131-2ubuntu2                                        amd64        Fast LZ compression algorithm library - development files
[estan@newton ~]$
```

**Info from the Ubuntu 17.10 system**

```
estan@balder-101:~$ dpkg -l | egrep "libblosc|liblz4"
ii  libblosc-dev                                  1.11.1+ds2-2                             amd64        high performance meta-compressor optimized for binary data (development files)
[...]

PR #216, opened 2018-02-15
[COMPAT] New split mode for forward compatibility by default

FrancescAlted:
New public function `blosc_set_splitmode()` for selecting another split mode than default.  The supported ones are:

 *  BLOSC_FORWARD_COMPAT_SPLIT (offers reasonably forward compatibility)
 *  BLOSC_AUTO_SPLIT (for nearly optimal results (based on heuristics))
 *  BLOSC_NEVER_SPLIT (for experimenting when trying to get compression ratios or speed)
 *  BLOSC_ALWAYS_SPLIT (for experimenting when trying to get compression ratios or speed)

Still need to implement support for setting the split mode via environment variables.


FrancescAlted:
Adresses #215.

kalvdans:
The patch looks good to me. Very nice to add tests for previous formats.

Should we add a note in the format description in `README_HEADER.rst` that compressor flags have changed meaning without bumping the compressor version? That will help other implementations also be forward compatible without having to dig into the git history.

I correct assuming the threadsafe API `blosc_compress_ctx()` will always use the reasonably forward compatibility mode? We should probably add a note to that function.

(I am a colleague to @estan)

FrancescAlted:
> The patch looks good to me. Very nice to add tests for previous formats.

The tests are not there yet (only the files), but my plan is to add them soon.

> Should we add a note in the format description in README_HEADER.rst that compressor flags have changed meaning without bumping the compressor version? That will help other implementations also be forward compatible without having to dig into the git history.

Yes, the plan is to add a section about this in the RELEASE_NOTES.rst.  However, the format has not changed this time; the change is only about having the split flag activated or not by default, so there should be not a need to change README_HEADER.rst.  The main message in the release notes should be: "from now on, all the buffers created starting with blosc 1.14.0 will be forward compatible with any previous versions of the library, at least until 1.3.0, when support for multi-codecs was introduced, and that of zstd codec is not used (zstd was introduced in 1.11.0).  Versions from 1.11.0 to 1.14.0 might generate buffers that cannot be read with versions < 1.11.0 (this was due to a forward incompatible change introduced in 1.11.1).". 

> Am I correct assuming the threadsafe API blosc_compress_ctx() will always use the reasonably forward compatibility mode? We should probably add a note to that function.

Yes.  The new 'BLOSC_FORWARD_COMPAT_SPLIT' mode will be the default for all the compression functions.  And the `blosc_set_splitmode()` function will affect to all compression functions too.

kalvdans:
> However, the format has not changed this time;

I was unclear, I meant the format change done in 9d06255. To avoid similar problems in future we should check that the reserved bits are zeros (I can make a pull request).

Your suggested release note sounds detailed and good!

FrancescAlted:
I see.  Agreed, I should have bumped the `BLOSC_VERSION_FORMAT` when I did
https://github.com/Blosc/c-blosc/commit/9d0625563a1472d0852b0328657b368723e4a95a.  Please feel free to file a PR about this.

estan:
This looks great.

Some comments:

I think the explanation of what `BLOSC_FORWARD_COMPAT_SPLIT` actually means should also go in the code documentation, not just the release notes.

But, I'm actually a little worried about the naming here. If I casually read `BLOSC_FORWARD_COMPAT_SPLIT` in some code, I'd assume that means forward compatibility is ensured, period. But when in fact, it's forward compat with a twist (not forward compatible if ZSTD is used).

Correct me if I'm wrong, but it is in fact the `BLOSC_ALWAYS_SPLIT` mode that ensures forward compatibility with all previous versions?

How about naming it like this:

 - `BLOSC_DEFAULT_SPLIT`
 - `BLOSC_OPTIMAL_SPLIT`
 - `BLOSC_NEVER_SPLIT`
 - `BLOSC_ALWAYS_SPLIT`
 - `BLOSC_FORWARD_COMPAT_SPLIT` (alias of `BLOSC_ALWAYS_SPLIT`)

The advantage is that when a user sees `BLOSC_DEFAULT_SPLIT`, will immediately know

1. It is the default.
2. The docs must be consulted to know what the meaning is (forward compat, with a little twist)

and when a user sees `BLOSC_FORWARD_COMPAT_SPLIT`, that is indeed (as can expected by the name) the mode that gives forward compatibility with all previous versions.

estan:
Sorry, my bad! I had missed the fact that zstd was introduced in the very same version as the no splitting stuff (I thought it was a couple of releases earlier). So the names do indeed make sense.

PR #217, opened 2018-02-16
Refuse to read future format versions

kalvdans:
Adds checks to header format version byte, to compressed data format
version byte, and that reserved fields are zero.

This change was triggered by the format version discussions in https://github.com/Blosc/c-blosc/pull/216 .
@FrancescAlted, @estan

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -91,7 +91,7 @@ struct blosc_context {
   const uint8_t* src;
   uint8_t* dest;                  /* The current pos in the destination buffer */
   uint8_t* header_flags;          /* Flags for header */
-  int versionlz;                  /* Compressor version byte, only used during decompression */
+  int compressor_version;         /* Compressor version byte, only used during decompression */
   int32_t sourcesize;             /* Number of bytes in source buffer (or uncompressed bytes in compressed file) */
   int32_t nblocks;                /* Number of total blocks in buffer */
   int32_t leftover;               /* Extra bytes at end of buffer */
@@ -689,7 +689,7 @@ static int blosc_d(struct blosc_context* context, int32_t blocksize,
   int doshuffle = (header_flags & BLOSC_DOSHUFFLE) && (typesize > 1);
   int dobitshuffle = ((header_flags & BLOSC_DOBITSHUFFLE) &&
                       (blocksize >= typesize));
-  int versionlz = context->versionlz;
+  int compressor_version = context->compressor_version;
 
   if (doshuffle || dobitshuffle) {
     _tmp = tmp;
@@ -718,21 +718,21 @@ static int blosc_d(struct blosc_context* context, int32_t blocksize,
     }
     else {
       if (compformat == BLOSC_BLOSCLZ_FORMAT) {
-        if (versionlz != BLOSC_BLOSCLZ_VERSION_FORMAT)
-          return -1;
+        if (compressor_version != BLOSC_BLOSCLZ_VERSION_FORMAT)
+          return -9;
         nbytes = blosclz_decompress(src, cbytes, _tmp, neblock);
       }
       #if defined(HAVE_LZ4)
       else if (compformat == BLOSC_LZ4_FORMAT) {
-        if (versionlz != BLOSC_LZ4_VERSION_FORMAT)
-          return -1;
+        if (compressor_version != BLOSC_LZ4_VERSION_FORMAT)
+          return -9;
         nbytes = lz4_wrap_decompress((char *)src, (size_t)cbytes,
                                      (char*)_tmp, (size_t)neblock);
       }
       #endif /*  HAVE_LZ4 */
       #if defined(HAVE_SNAPPY)
       else if (compformat == BLOSC_SNAPPY_FORMAT) {
-        if (versionlz != BLOSC_SNAPPY_VERSION_FORMAT)
+        if (compressor_version != BLOSC_SNAPPY_VERSION_FORMAT)
           return -1;
         nbytes = snappy_wrap_decompress((char *)src, (size_t)cbytes,
                                         (char*)_tmp, (size_t)neblock);
@@ -740,7 +740,7 @@ static int blosc_d(struct blosc_context* context, int32_t blocksize,
       #endif /*  HAVE_SNAPPY */
       #if defined(HAVE_ZLIB)
       else if (compformat == BLOSC_ZLIB_FORMAT) {
-        if (versionlz != BLOSC_ZLIB_VERSION_FORMAT)
+        if (compressor_version != BLOSC_ZLIB_VERSION_FORMAT)
           return -1;
         nbytes = zlib_wrap_decompress((char *)src, (size_t)cbytes,
                                       (char*)_tmp, (size_t)neblock);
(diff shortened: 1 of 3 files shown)


blosc/blosc.c
```diff
@@ -1355,18 +1364,21 @@ int blosc_run_decompression_with_context(struct blosc_context* context,
...
   context->sourcesize = sw32_(context->src + 4);     /* buffer size */
   context->blocksize = sw32_(context->src + 8);      /* block size */
-  ctbytes = sw32_(context->src + 12);               /* compressed buffer size */
 
-  /* Unused values */
-  version += 0;                             /* shut up compiler warning */
-  versionlz += 0;                           /* shut up compiler warning */
-  ctbytes += 0;                             /* shut up compiler warning */
+  if (version != BLOSC_VERSION_FORMAT) {
+    /* Version from future */
```
FrancescAlted: May it be worth to print some informative message before giving up?  Something like:

```
printf("Unrecognized header version %d\n", version);
```

blosc/blosc.c
```diff
@@ -1355,18 +1364,21 @@ int blosc_run_decompression_with_context(struct blosc_context* context,
...
+    return -1;
+  }
+  if (*context->header_flags & 0x08) {
+    /* compressor flags from the future */
+    return -1;
```
FrancescAlted: Ditto.

blosc/blosc.c
```diff
@@ -1540,10 +1552,11 @@ int blosc_getitem(const void *src, int start, int nitems, void *dest)
       cbytes = bsize2;
     }
     else {
-      struct blosc_context context;
-      /* blosc_d only uses typesize and flags */
+      struct blosc_context context = {0};
```
FrancescAlted: +1 Nice way to initialize struct to 0.

blosc/blosc.c
```diff
@@ -1969,14 +1982,9 @@ void blosc_cbuffer_sizes(const void *cbuffer, size_t *nbytes,
...
-  uint8_t version, versionlz;              /* versions for compressed header */
-
-  /* Read the version info (could be useful in the future) */
-  version = _src[0];                       /* blosc format version */
-  versionlz = _src[1];                     /* blosclz format version */
+  uint8_t version = _src[0];               /* version of header */
 
-  version += 0;                            /* shut up compiler warning */
-  versionlz += 0;                          /* shut up compiler warning */
+  assert (version == BLOSC_VERSION_FORMAT); /* Should have been checked earlier */
```
FrancescAlted: Is this comment applicable to a code that is going to be included in the library?  As it is I find it misleading.

blosc/blosc.c
```diff
@@ -1990,14 +1998,8 @@ void blosc_cbuffer_metainfo(const void *cbuffer, size_t *typesize,
...
   uint8_t *_src = (uint8_t *)(cbuffer);  /* current pos for source buffer */
-  uint8_t version, versionlz;            /* versions for compressed header */
-
-  /* Read the version info (could be useful in the future) */
-  version = _src[0];                     /* blosc format version */
-  versionlz = _src[1];                   /* blosclz format version */
 
-  version += 0;                             /* shut up compiler warning */
-  versionlz += 0;                           /* shut up compiler warning */
+  assert (_src[0] == BLOSC_VERSION_FORMAT); /* Should have been checked earlier */
```
FrancescAlted: Ditto.

FrancescAlted:
I generally agree with the PR.  My only suggestion is to add some explanatory messages for the user understanding the reason why a Blosc buffer cannot be decoded.

blosc/blosc.c
kalvdans: I don't like libraries that print messages, that descision should be given to the caller. Before we have a way to return error messages, I prefer silent errors. Maybe we can return -10 and document what it means in the API documentation?

blosc/blosc.c
kalvdans: Yes, I agree the comment is misleading, I will remove the comment. Since there is no way to signal errors I'd like to keep the assert though.

blosc/blosc.c
```diff
@@ -90,10 +90,8 @@ struct blosc_context {
 
   const uint8_t* src;
   uint8_t* dest;                  /* The current pos in the destination buffer */
-  uint8_t* header_flags;          /* Flags for header.  Currently booked:
-                                    - 0: byte-shuffled?
-                                    - 1: memcpy'ed?
-                                    - 2: bit-shuffled? */
+  uint8_t* header_flags;          /* Flags for header */
+  int versionlz;                  /* Compressor version byte, only used during decompression */
```
estan: How about calling it just `compressor_version`? (It looks like it's used for more than just LZ-based compressors?)

blosc/blosc.c
kalvdans: Thanks! It is makes the code slightly slower, but so much easier to debug!

kalvdans:
@FrancescAlted, @estan: I have made version mismatch return a special value and documented that value, and renamed the variable to be algorithm-agnostic. Please review again! Also I don't know how to debug the failed test on osx.

estan:
LGTM. I don't know about the macOS failure, but maybe it's some lingering effect of this incident? https://www.traviscistatus.com/incidents/t68jk6rj92q4

FrancescAlted:
Yep, I restored the failing travis job and everything passes smoothly now.

blosc/blosc.c
FrancescAlted: I see.  However, the current approach is to print an informative message before returning a negative number.  I don't think that changing this behavior in this PR in just a couple of places is a good idea.  Probably the good thing to do is to detect all the places where a negative code is returned, make sure that there are no repetitions and document these codes.  One can even think on creating a new API (`blosc_print_error(int)?) for more verbose output.  But this would be another PR.

blosc/blosc.c
```diff
@@ -1455,7 +1467,7 @@ int blosc_decompress(const void *src, void *dest, size_t destsize)
 int blosc_getitem(const void *src, int start, int nitems, void *dest)
 {
   uint8_t *_src=NULL;               /* current pos for source buffer */
-  uint8_t version, versionlz;       /* versions for compressed header */
+  uint8_t version, compressor_version; /* versions for compressed header */
```
FrancescAlted: For being consistent with blosc convention for names(e.g. `compname`, `compformat`), I'd prefer using `compversion` for this variable.

blosc/blosc.c
kalvdans: Yes, we need at least to move the error printing out of the helper threads to avoid printing garbled output. I can do that refactor in a separate pull request with the existing error messages, and push that first, if you want.

FrancescAlted: Well, maybe a printf is not strictly necessary.  Perhaps a `char* blosc_get_error(int)` would be enough.  A better approach would be to define an enumerated for the different errors and some functions for returning proper messages.  The [Zstandard approach](https://github.com/facebook/zstd/blob/dev/lib/common/zstd_common.c#L31) could be a good example to follow.

estan: That sounds like a good idea to me, the Zstd approach resembles what I've seen in other libraries. I also don't think a library should print things unless told to - what to do with the error should be the user's decision (where the user here is a programmer).

FrancescAlted:
Superseded by https://github.com/Blosc/c-blosc/pull/219

PR #218, opened 2018-02-17
Turn on useful warnings from gcc

kalvdans:
Fix code that gcc warns for. Remove unused variables, const poisoning,
and use declare functions (void) instead of ().

Please note that it changes the signatures of the API. The ABI is not changed.

I need someone to double-check the CMakeLists.txt changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,7 +253,7 @@ endif()
 # flags
 # set -Wall and other useful warning flags
 if(CMAKE_C_COMPILER_ID STREQUAL GNU OR CMAKE_C_COMPILER_ID STREQUAL Clang OR CMAKE_C_COMPILER_ID STREQUAL Intel)
-     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wwrite-strings -Wstrict-prototypes")
+     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wwrite-strings -Wstrict-prototypes -Wno-unused-function")
 endif(CMAKE_C_COMPILER_ID STREQUAL GNU OR CMAKE_C_COMPILER_ID STREQUAL Clang OR CMAKE_C_COMPILER_ID STREQUAL Intel)
 
 # @NOTE: -O3 is enabled in Release mode (CMAKE_BUILD_TYPE="Release")
@@ -262,12 +262,12 @@ endif(CMAKE_C_COMPILER_ID STREQUAL GNU OR CMAKE_C_COMPILER_ID STREQUAL Clang OR 
 # Probably "-msse2" should be appended to CMAKE_C_FLAGS_RELEASE.
 if(CMAKE_C_COMPILER_ID STREQUAL GNU OR CMAKE_C_COMPILER_ID STREQUAL Clang OR CMAKE_C_COMPILER_ID STREQUAL Intel)
      if(COMPILER_SUPPORT_SSE2)
-         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse2" CACHE STRING "C flags." FORCE)
+         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse2")
      endif(COMPILER_SUPPORT_SSE2)
 endif(CMAKE_C_COMPILER_ID STREQUAL GNU OR CMAKE_C_COMPILER_ID STREQUAL Clang OR CMAKE_C_COMPILER_ID STREQUAL Intel)
 
 if(MSVC)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Ox" CACHE STRING "C flags." FORCE)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Ox")
 
     # Turn off misguided "secure CRT" warnings in MSVC.
     # Microsoft wants people to use the MS-specific <function>_s
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,23 +251,22 @@ if(DEACTIVATE_AVX2)
 endif()
 
 # flags
-# set -Wall and other useful warning flags
+# Set -Wall and other useful warning flags.
 if(CMAKE_C_COMPILER_ID STREQUAL GNU OR CMAKE_C_COMPILER_ID STREQUAL Clang OR CMAKE_C_COMPILER_ID STREQUAL Intel)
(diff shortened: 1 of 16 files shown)


kalvdans:
@FrancescAlted, what is your opinion on annotating strings that should not be modified with 'const' in the API?

kalvdans:
Ok, this pull request is ready for review. @estan, @FrancescAlted. We also need to agree on the timing to introduce these API changes and what to write in the release notes.

CMakeLists.txt
```diff
@@ -251,15 +251,18 @@ if(DEACTIVATE_AVX2)
...
-# Set the "-msse2" build flag only if the CMAKE_C_FLAGS is not already set.
-# Probably "-msse2" should be appended to CMAKE_C_FLAGS_RELEASE.
+# Set the "-msse2" build flag if supported.
 if(CMAKE_C_COMPILER_ID STREQUAL GNU OR CMAKE_C_COMPILER_ID STREQUAL Clang OR CMAKE_C_COMPILER_ID STREQUAL Intel)
-     if(NOT CMAKE_C_FLAGS AND COMPILER_SUPPORT_SSE2)
-         set(CMAKE_C_FLAGS -msse2 CACHE STRING "C flags." FORCE)
-     endif(NOT CMAKE_C_FLAGS AND COMPILER_SUPPORT_SSE2)
+     if(COMPILER_SUPPORT_SSE2)
+         add_compile_options(-msse2)
+     endif(COMPILER_SUPPORT_SSE2)
```
FrancescAlted: Does add_compile_options() require a cmake > 2.8.10?  This is the current minimum required.

blosc/blosclz.c
```diff
@@ -79,7 +79,7 @@
 
 
 
-static inline uint8_t *get_run(uint8_t *ip, const uint8_t *ip_bound, const uint8_t *ref) {
+static uint8_t *get_run(uint8_t *ip, const uint8_t *ip_bound, const uint8_t *ref) {
```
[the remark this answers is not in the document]
kalvdans: This change was unintentional. I will revert it. And yes, modern compilers ignore this keyword, I usually use `__attribute__((force_inline))` instead.

tests/test_shuffle_roundtrip_avx2.c
```diff
@@ -18,12 +18,6 @@
 
 #if defined(SHUFFLE_AVX2_ENABLED)
   #include "../blosc/shuffle-avx2.h"
-#else
-  #if defined(_MSC_VER)
-  #pragma message("AVX2 shuffle tests not enabled.")
-  #else
-  #warning AVX2 shuffle tests not enabled.
-  #endif
```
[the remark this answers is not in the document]
kalvdans: It breaks the build if we want to enable -Werror at a later time. Do you feel strongly for keeping it?

CMakeLists.txt
kalvdans: Good catch, seems it was added in cmake 2.8.12. Should we require cmake >= 2.8.12 or should I revert my changes to `CMakeLists.txt`?

tests/test_shuffle_roundtrip_avx2.c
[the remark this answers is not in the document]
FrancescAlted: Well, I am just used to see it.  If I see it, I know that AVX2 is not being tested in the platform.  Probably mainly useful during early stages of AVX2 support though, so +1.

CMakeLists.txt
FrancescAlted: Fortunately it seems like [Ubuntu trusty has 2.8.12](https://packages.ubuntu.com/trusty/cmake), so let's make a bump.  Could you do that in main CMakeLists.txt?

CMakeLists.txt
```diff
@@ -251,15 +251,18 @@ if(DEACTIVATE_AVX2)
 endif()
 
 # flags
-# @TODO: set -Wall
+# Set -Wall and other useful warning flags.
+if(CMAKE_C_COMPILER_ID STREQUAL GNU OR CMAKE_C_COMPILER_ID STREQUAL Clang OR CMAKE_C_COMPILER_ID STREQUAL Intel)
```
estan: You could also toss `AppleClang` in there for good measure. It's how Apple's own version of Clang is identified.

Instead of the `if`, you could use generator expressions like:

```cmake
add_compile_options(
    "$<$<CXX_COMPILER_ID:GNU>:-Wall;-Wwrite-strings;-Wno-unused>"
    "$<$<CXX_COMPILER_ID:Clang>:-Wall;-Wwrite-strings;-Wno-unused>"
    "$<$<CXX_COMPILER_ID:AppleClang>:-Wall;-Wwrite-strings;-Wno-unused>"
    "$<$<CXX_COMPILER_ID:Intel>:-Wall;-Wwrite-strings;-Wno-unused>"
)
```

But I don't know if that's any prettier? Best-looking would be if the 3.3+ feature `if ("bar" IN_LIST _list)` could be used I think, but that's not in 2.8.

If you keep it the way it is, perhaps move the `add_compile_option` to inside the `if` block just below (the one for SSE2)? (since the conditions are the same).

CMakeLists.txt
```diff
@@ -251,15 +251,18 @@ if(DEACTIVATE_AVX2)
...
+    add_compile_options(-Wall -Wwrite-strings -Wno-unused-function)
+endif(CMAKE_C_COMPILER_ID STREQUAL GNU OR CMAKE_C_COMPILER_ID STREQUAL Clang OR CMAKE_C_COMPILER_ID STREQUAL Intel)
```
estan: Repeating the condition in the `endif(..)` is no longer necessary I think, so this could be just `endif()`.

FrancescAlted: Repeating the condition in the `endif(...)` was just a way to see at a glance where the condition block finished.  This is more useful in nested conditions, and we could remove it from here (not a big deal though).

estan:
@kalvdans Looks like some nice cleanups. Just a couple of comments, but :+1:  from me.

CMakeLists.txt
estan: In fact, thinking a bit, don't take the generator expressions as a serious suggestion. It's just cryptic, while anyone who reads the `CMakeLists.txt` probably just want to get on with what they're doing and not learn all about CMake's warts :)

Generator expressions seems to be primarily for when you need information that is only available at CMake's final "generation" step, and the compiler ID is not such a thing.

CMakeLists.txt
estan: @FrancescAlted Aha, I thought it was a left-over from < 2.5 times, when it was mandatory. Yes, no big deal, and could just as well be kept.

PR #221, opened 2018-03-04
Declare fsize at the beginning on main

jakirkham:
Moves the `fsize` type definition to the top of `main` to comply with C89's constraints. Should fix a compilation error on VS 2008 (needed for Python 2.7 support on Windows).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,6 +15,9 @@ environment:
     matrix:
         - MINGW_CONFIGURATIONS: "4.9@x86_64@seh@posix, 5@x86_64@seh@posix, 6@x86_64@seh@posix, 7@x86_64@seh@posix"
         - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+          CONAN_VISUAL_VERSIONS: 9
+          CONAN_RUN_TESTS: "1"
+        - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
           CONAN_VISUAL_VERSIONS: 12
           CONAN_RUN_TESTS: "1"
         - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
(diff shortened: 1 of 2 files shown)


jakirkham:
Any thoughts on this @FrancescAlted?

AFAICT the macOS build failures are unrelated as they are failing well before the build begins (some issue installing things on Homebrew).

kalvdans:
If C89 is required we should use the -std=c89 flag to gcc on Linux as well, to discover mistakes locally before the travis build.

jakirkham:
SGTM.

Though the issue occurred on Windows. So it might be worth adding a VS 2008 build there. For simplicity would go with VS 2008 32-bit. Trying to get 64-bit to work requires a hack. We can do it easily on AppVeyor, but it might just be simpler to skip it.

Will wait a little bit and see what others think on this testing issue.

FrancescAlted:
I am in vacation now. I'll look into this by the end of the week.

FrancescAlted:
The patch looks good toi me.  And yes, I agree that adding the -std=c89 flag for gcc should be enough to catch these.  However, adding a new configuration for VS2008 would work best and I personally prefer this.  Could you please add a new config for VS2008?  Also, any hint on why the OSX tests are not working?  Is this a temporary issue with Travis?

jakirkham:
Sure, I can take a look. Might not be until next week though.

The error on macOS is Homebrew related. Some [issue installing Python 3]( https://travis-ci.org/Blosc/c-blosc/jobs/348856351#L2010-L2015 ). Unfortunately I don't have more time to sink into that.

FrancescAlted:
Great.  Thanks.

jakirkham:
So I tried to add a line to use VS 2008. Unfortunately it doesn't appear to be working and am really not sure why.

FrancescAlted:
Ok.  So, don't know what's happening with VS2008, but I have moved the `fsize` declaration.  I'm still open for suggestions of how to enforce C89 for C-Blosc (C-Blosc2 will be C99 for sure).

jakirkham:
Thanks for picking up this change.

I thinks @kalvdans suggestion might be the best way to go for now. At least that would allow for C89 testing, which should be in practice sufficient even if it is not the same as testing on VS 2008 itself.

PR #224, opened 2018-04-05
always use win32 threads on Windows

stevengj:
On Windows with non-GNU compilers, `blosc.c` uses `win32/pthread.c` as a lightweight pthreads replacement using native Windows threads.

This PR changes it to use `win32/pthread.c` on **all Windows builds** even with GNU compilers.   Although MinGW etcetera provides a more full-featured pthreads replacement, it doesn't seem to accomplish anything here since the functionality in `win32/pthread.c` is sufficient for Blosc.  Furthermore, using the MinGW pthreads adds an additional library dependency to libblosc that is annoying for binary distribution.  (For example, it got in the way of distributing cross-compiled Windows binaries for use with Julia, since we want the resulting `libblosc.dll` to be usable on any Windows machine even where MinGW is not installed).

I've been patching c-blosc locally with this change, but it doesn't seem like there is any reason not to make the change globally.

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -56,7 +56,7 @@
   #include <inttypes.h>
 #endif  /* _WIN32 */
 
-#if defined(_WIN32) && !defined(__GNUC__)
+#if defined(_WIN32)
   #include "win32/pthread.h"
   #include "win32/pthread.c"
 #else


stevengj:
Thanks!

By the way, I've set up automatic cross-compilation of Blosc binaries for multiple platforms at https://github.com/stevengj/BloscBuilder ... if you look at the [releases](https://github.com/stevengj/BloscBuilder/releases) page, you'll see releases with tarballs attached that provide cross-compiled shared libraries for multiple platforms.   (This makes it easy for Julia users to install without being dependent on a local build system.)   Don't know if this is useful for `python-blosc` or similar packages?

FrancescAlted:
Thanks for the heads ups Steven.  Yeah, definitely useful, specially for [pycblosc](https://github.com/Blosc/pycblosc), a very light-weight C-Blosc wrapper that I have in the works.  This is currently using [Conan](https://bintray.com/blosc/Conan), a service for hosting binaries.  I am experimenting with Conan yet, and still need some more time with it, but good to know that you have solved the problem of binaries differently.  May I ask if there is some automated way to rebuild the binaries when a new C-Blosc release is out?  Some recommendation for other packagers to use your binary repo?

stevengj:
@FrancescAlted, yes, to build a new c-blosc release I would just need to update the [c-blosc commit tag in the build_tarballs.jl](https://github.com/stevengj/BloscBuilder/blob/fce53a562e238a377fd86bf7d7d74fe75d0342bd/build_tarballs.jl#L6) script and tag a new release.

Whenever a release of BloscBuilder is tagged, Travis is automatically triggered to run a series of cross-compilers and upload the resulting binaries to the release.  (This is all thanks to the Julia [BinaryBuilder package](https://github.com/JuliaPackaging/BinaryBuilder.jl).)

For other packagers to use the BloscBuilder binaries, they just need to download a [`.tar.gz` file attached to a release](https://github.com/stevengj/BloscBuilder/releases) and unpack it.  This gives them a `libblosc` shared library for whatever platform they downloaded.

PR #241, opened 2018-10-21
Enable fork-safe operation via pthread_atfork hook.

asford:
The blosc global context object maintains a thread-pool and internal
mutexs, and is unsafe to reused in a child process post fork. This
causes deadlocks in child processes post-fork, which affects any
blosc-consuming libraries. The `_ctx` context apis (set_releasegil in
python-blosc) provide a workaround, but have a non-zero overhead and are
non-default. (https://github.com/Blosc/bloscpack/issues/58)

This update provides fork-safety by discarding and reinitializing the
global context in the post-fork child. The behavior "leaks" the previous
context in the child by skipping cleanup, however the context object is
already in an invalid state and likely can't be reliably reused. 

* Add init with pthread_atfork hook.

  Update blosc to reinitialize child global context post-fork,
  discarding potentially inconsistent global context. Resolves issues with
  deadlock-on-decompression in child process post fork if blosc has
  already been initialized.

* Add test for compressor fork-safety.

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -2117,6 +2117,20 @@ void blosc_set_splitmode(int mode)
   g_splitmode = mode;
 }
 
+/* Child process global context is an invalid state and threads are no longer
+ * valid.  Re-init a new context. */
+void blosc_fork_child() {
+  if (!g_initlib) return;
+
+  /* trylock, fork may have left global mutex locked in another thread */
+  pthread_mutex_trylock(&global_comp_mutex);
+
+  g_global_context = (struct blosc_context*)my_malloc(sizeof(struct blosc_context));
+  g_global_context->threads_started = 0;
+
+  pthread_mutex_unlock(&global_comp_mutex);
+}
+
 void blosc_init(void)
 {
   /* Return if we are already initialized */
@@ -2125,6 +2139,11 @@ void blosc_init(void)
   pthread_mutex_init(&global_comp_mutex, NULL);
   g_global_context = (struct blosc_context*)my_malloc(sizeof(struct blosc_context));
   g_global_context->threads_started = 0;
+
+  #if !defined(_WIN32)
+      pthread_atfork(NULL, NULL, &blosc_fork_child);
+  #endif
+
   g_initlib = 1;
 }
 
--- a/tests/test_forksafe.c
+++ b/tests/test_forksafe.c
@@ -1,0 +1,104 @@
+/*********************************************************************
+  Blosc - Blocked Shuffling and Compression Library
+
+  Roundtrip compression/decompression tests.
+
+  Creation date: 2010-06-07
+  Author: Francesc Alted <francesc@blosc.org>
+
+  See LICENSES/BLOSC.txt for details about copyright and rights to use.
+**********************************************************************/
+
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <signal.h>
+
+#include "test_common.h"
+
+
+int clevel = 3;
+int doshuffle = 1;
+size_t typesize = 4;
+size_t size = 5*MB;
+#define BUFFER_ALIGN_SIZE   8
+
+void *src, *dest, *dest2;
+size_t nbytes, cbytes;
+
+int tests_run = 0;
+
+static char *test_forksafe(void) {
+  /* Compress the input data and store it in dest. */
+  blosc_set_nthreads(4);
+  cbytes = blosc_compress(clevel, doshuffle, typesize, size, src, dest, size);
+
+  pid_t newpid = fork();
+  if (newpid == 0) {
+    nbytes = blosc_decompress(dest, dest2, size);
+    exit(0);
+  }
+
+  int success = 0;
+	int status;
+  for(float sec = 0; sec < 1; ) {
+    if(waitpid(newpid, &status, WNOHANG) != 0) {
+      success = 1;
+      break;
+    }
+
+    usleep(5000);
+    sec += 5000 * 1e-6;
+  }
+
+  if(!success) {
+    kill(newpid, SIGKILL);
+    waitpid(newpid, &status, 0);
+  }
+
+  mu_assert("ERROR: Child deadlocked post-fork.", success == 1);
+
+  return 0;
+}
+
+
+static const char *all_tests(void) {
+  mu_run_test(test_forksafe);
+
+  return 0;
+}
+
+int main(int argc, char **argv) {
+  const char *result;
+
+  printf("STARTING TESTS for %s", argv[0]);
+
+  blosc_init();
+
+  /* Initialize buffers */
+  src = blosc_test_malloc(BUFFER_ALIGN_SIZE, size);
+  dest = blosc_test_malloc(BUFFER_ALIGN_SIZE, size + 16);
+  dest2 = blosc_test_malloc(BUFFER_ALIGN_SIZE, size);
+
+  /* Fill the input data buffer with random values. */
+  blosc_test_fill_random(src, size);
+
+  /* Run all the suite */
+  result = all_tests();
+  if (result != 0) {
+    printf(" (%s)\n", result);
+  }
+  else {
+    printf(" ALL TESTS PASSED");
+  }
+  printf("\tTests run: %d\n", tests_run);
+
+  blosc_test_free(src);
+  blosc_test_free(dest);
+  blosc_test_free(dest2);
+
+  blosc_destroy();
+
+  return result != 0;
+}
--- a/blosc/win32/pthread.h
+++ b/blosc/win32/pthread.h
@@ -45,6 +45,7 @@
 #define pthread_mutex_init(a,b) InitializeCriticalSection((a))
 #define pthread_mutex_destroy(a) DeleteCriticalSection((a))
 #define pthread_mutex_lock EnterCriticalSection
+#define pthread_mutex_trylock TryEnterCriticalSection
 #define pthread_mutex_unlock LeaveCriticalSection
 
 /*


asford:
## python-blosc exhaustive repro

```
import logging
import functools

import os
import time
import sys

import blosc

logging.basicConfig(
    level=logging.INFO,
    format="%(asctime)s.%(msecs).03d pid: %(process)d %(message)s",
    datefmt='%Y-%m-%dT%H:%M:%S')


def log(func):
    @functools.wraps(func)
    def wfunc(*args, **kwargs):
        try:
            logging.info("enter: %s" % func.__name__)
            return func(*args, **kwargs)
        finally:
            logging.info("exit: %s" % func.__name__)

    return wfunc


def waitfor(pid, timeout=0.0):
    step = .05
    if timeout is None:
        return os.waitpid(pid, 0)
    else:
        assert timeout > 0
        left = timeout
        while left > 0:
            result = os.waitpid(pid, os.WNOHANG)
            if result != (0, 0):
                return result
            time.sleep(step)
            left -= step
        else:
            raise TimeoutError(
                "process did not exit pid: %s timeout: %s" % (pid, timeout))


@log
def load_buf(buf):
    blosc.decompress(buf)


@log
def load_buf_in_fork(buf):
    newpid = os.fork()
    if newpid == 0:
        load_buf(buf)
        sys.exit()
    waitfor(newpid, 2.0)


@log
def init_buf_in_fork(size):
    newpid = os.fork()
    if newpid == 0:
        with open("buf.blosc", "wb") as o:
            o.write(blosc.compress(os.urandom(size)))
        sys.exit()
    waitfor(newpid, 2.0)

    try:
        with open("buf.blosc", "rb") as i:
            return i.read()
    finally:
        os.remove("buf.blosc")


@log
def main():
    buf = init_buf_in_fork(5 * 1024 * 1024)

    logging.info('decompress in fork process pre-init OK')
    load_buf_in_fork(buf)

    logging.info('decompress in main process OK')
    load_buf(buf)

    logging.info('decompress in fork with gilrelease OK')
    blosc.set_releasegil(True)
    load_buf_in_fork(buf)

    logging.info('decompress in fork after decompress in main process hangs')
    blosc.set_releasegil(False)
    load_buf_in_fork(buf)


if __name__ == '__main__':
    main()
```

blosc/blosc.c
```diff
@@ -2117,6 +2117,20 @@ void blosc_set_splitmode(int mode)
...
+void blosc_fork_child() {
+  if (!g_initlib) return;
+
+  /* trylock, fork may have left global mutex locked in another thread */
+  pthread_mutex_trylock(&global_comp_mutex);
+
+  g_global_context = (struct blosc_context*)my_malloc(sizeof(struct blosc_context));
+  g_global_context->threads_started = 0;
+
+  pthread_mutex_unlock(&global_comp_mutex);
```
kalvdans: You cannot unlock a mutex that you don't own. I think you have to reinitialize the lock instead.

blosc/blosc.c
```diff
@@ -2117,6 +2117,20 @@ void blosc_set_splitmode(int mode)
   g_splitmode = mode;
 }
 
+/* Child process global context is an invalid state and threads are no longer
+ * valid.  Re-init a new context. */
+void blosc_fork_child() {
```
kalvdans: `void blosc_fork_child(void)`

FrancescAlted:
@asford Thanks for the PR.  I am interested in applying it.  It is unfortunate that the CI test are undergoing some problem (apparently related with Conan packages).

@kalvdans I need to to think a bit more on your suggestion about "remove global_comp_mutex" and see if that has side effects.  IIUC, you think this would render this PR unnecessary?

kalvdans:
> @kalvdans I need to to think a bit more on your suggestion about "remove global_comp_mutex" and see if that has side effects. IIUC, you think this would render this PR unnecessary?

I thought so initially, but at a closer thought it is still necessary as long as the thread pool is active at the time of fork(). After all, @asford 's test case is single-threaded.

blosc/blosc.c
asford: Thank you for the review. This uncovered an error causing in duplicate atfork handler registration, which I've resolved.

asford:
🤔 I would prefer to maintain the `global_comp_mutex` used in the library. You can then round the library up to "thread-safe & fork-safe", rather than trading thread-safety for fork-safety, and (as @kalvdans mentioned) fork-safety will require `atfork` handlers in either case.

Also, though c-blosc is _often_ accessed through a GIL-protected python interface, there may be dependents that rely on the existing thread-safe behavior. Shifting from thread-safe to non-thread-safe is a potentially breaking, but isn't a _loud_ build-breaking API change and may introduce hard to detect/test bugs in downstream consumers.

asford:
@FrancescAlted It appears that there was some form of breaking change in `conan_package_tools` in the `0.19.[4-7]` release series. I've no familiarity with the tool, but pinning to the version used in your last `master` CI run appears to resolve the issue.

Should we keep the version pin in this PR, or would you prefer to break it out?

blosc/blosc.c
```diff
@@ -2117,6 +2118,34 @@ void blosc_set_splitmode(int mode)
   g_splitmode = mode;
 }
 
+/* Ensure global context safety post-fork via:
+ * 1. Acquire global context mutex pre-fork, blocks fork until global context
+ *    in released in other threads.
```
kalvdans: Very good to add lots of comments to this complex problem!

Waiting for the lock pre-fork will cause deadlock in situations where a blosc worker thread wants to do a fork-exec. Maybe a bit exotic but a simple system("echo I was here >> /tmp/debug.log"); will cause deadlock. I am fine with that as long as we are aware of the consequences. It doesn't hurt to add error checking to the pthread_mutex_lock calls, since in some cases the deadlock can be discovered and EDEADLK be returned.

tests/test_forksafe.c
```diff
@@ -0,0 +1,103 @@
+/*********************************************************************
+  Blosc - Blocked Shuffling and Compression Library
+
+  Roundtrip compression/decompression tests.
+
+  Creation date: 2010-06-07
+  Author: Francesc Alted <francesc@blosc.org>
```
kalvdans: Cut-and-paste accident in the Author and summary.

blosc/blosc.c
asford: I believe the most-portable solution here is to convert `global_comp_mutex` to a `pthread_mutex_t*`, then fully de-initialize the global context in the child post-fork. Subsequent calls to the `blosc_*` interfaces will trigger re-init of the global context.

blosc/blosc.c
asford: This potential deadlock is resolved by the lazy-reinit strategy described in the alternate comment thread. Instead, it will become unsafe to `fork` while holding the `global_comp_mutex` and continue execution from the child process. As this mutex is an internal detail of `blosc.c`, and blosc isn't going to be using a fork-based pool strategy, this scenario shouldn't occur.

blosc/blosc.c
kalvdans: Very interesting read, the history of phtread_atfork! Not only does POSIX forbid calling async-signal-unsafe functions in post-fork handler, but also during the whole lifetime of the child between fork and exec. Which means your test case is invalid code since blosc API isn't async-singnal-safe.

kalvdans: Yes, it makes sense. Let's go on with this patch.

tests/test_forksafe.c
```diff
@@ -0,0 +1,103 @@
+/*********************************************************************
+  Blosc - Blocked Shuffling and Compression Library
+
+  Roundtrip compression/decompression tests.
```
kalvdans: The summary is still wrong. I suggest "Tests that the library can be used after fork()"

tests/test_forksafe.c
```diff
@@ -0,0 +1,103 @@
...
+    usleep(5000);
+    sec += 5000 * 1e-6;
+  }
+
+  if(!success) {
+    kill(newpid, SIGKILL);
+    waitpid(newpid, &status, 0);
+  }
+
+  mu_assert("ERROR: Child deadlocked post-fork.", success == 1);
```
kalvdans: `mu_assert("ERROR: Child crashed.", status == 0);`

tests/test_forksafe.c
```diff
@@ -0,0 +1,103 @@
...
+int tests_run = 0;
+
+static char *test_forksafe(void) {
+  /* Compress the input data and store it in dest. */
+  blosc_set_nthreads(4);
+  cbytes = blosc_compress(clevel, doshuffle, typesize, size, src, dest, size);
+
+  pid_t newpid = fork();
+  if (newpid == 0) {
+    nbytes = blosc_decompress(dest, dest2, size);
```
kalvdans: `assert (nbytes == size);`

blosc/blosc.c
```diff
@@ -2117,14 +2118,42 @@ void blosc_set_splitmode(int mode)
...
+ * mark as uninitialized.  Subsequent calls through `blosc_*` interfaces will
+ * trigger re-init of the global context.
+ *
+ * All pthread interfaces have undefined behavior in child handler in current
+ * posix standards: http://pubs.opengroup.org/onlinepubs/9699919799/
+ */
+void blosc_atfork_child(void) {
+  if (!g_initlib) return;
+
+  global_comp_mutex = NULL;
```
kalvdans: I believe we could do `pthread_mutex_init()` here instead, and not use a pointer. But I'm fine either way.

tests/test_forksafe.c
asford:
```suggestion
  Test that library can be used after fork().
```

tests/test_forksafe.c
asford:
```suggestion
    nbytes = blosc_decompress(dest, dest2, size);
    mu_assert(nbytes == size);
```

tests/test_forksafe.c
asford:
```suggestion
  mu_assert("ERROR: Child deadlocked post-fork.", success == 1);
  mu_assert("ERROR: Child crashed.", status == 0);
```

kalvdans:
Looks good to me now!

> Shifting from thread-safe to non-thread-safe is a potentially breaking, but isn't a _loud_ build-breaking API change and may introduce hard to detect/test bugs in downstream consumers.

Very good point. Silent API changes are the worst.

blosc/blosc.c
asford: Just a note, but a call to `pthread_mutex_init` on a previously-init-ed mutex location without a successful `pthread_mutex_destroy` [results in undefined behavior](http://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_mutex_init.html). I believe that some form of pointer is required so that the value can be abandoned post-fork.

blosc/blosc.c
```diff
@@ -2136,7 +2165,7 @@ void blosc_destroy(void)
   g_initlib = 0;
   blosc_release_threadpool(g_global_context);
   my_free(g_global_context);
-  pthread_mutex_destroy(&global_comp_mutex);
+  pthread_mutex_destroy(global_comp_mutex);
```
FrancescAlted: Should we require a `free(global_comp_mutex)` here?

FrancescAlted: Also, I do not get why we need to create the `global_comp_mutex` in the heap, but I suppose you have good reasons for this.

blosc/blosc.c
FrancescAlted: Should we use `free(global_comp_mutex)` previous to this?

FrancescAlted:
Good job @asford and @kalvdans !  Other than a possible need of `free()` in a couple of place of spots (see my comments above), I am +1 on this. @asford would you mind to add an entry about this PR in the RELEASE_NOTES.rst file?

Also, good that you figured out how to fix the CI issues with Conan.  Frankly, although it is nice to have the possibility to create binaries for Blosc automatically, I am not so happy with Conan because of this fragility when updating new Conan versions.  Your pinning seems a good workaround for now though :)

PR #244, opened 2018-11-05
Small fixes

AndreasMartin72:
- Changed inline modifiers (compiler dependent)
- Replaced //-comments with /**/-comments

### DIFF
--- a/blosc/blosc-comp-features.h
+++ b/blosc/blosc-comp-features.h
@@ -1,0 +1,23 @@
+/*********************************************************************
+  Blosc - Blocked Shuffling and Compression Library
+
+  Author: Francesc Alted <francesc@blosc.org>
+
+  See LICENSES/BLOSC.txt for details about copyright and rights to use.
+**********************************************************************/
+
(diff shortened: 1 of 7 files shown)


blosc/blosc-comp-features.h
```diff
@@ -0,0 +1,23 @@
...
+**********************************************************************/
+
+#ifndef BLOSC_COMP_FEATURES_H
+#define BLOSC_COMP_FEATURES_H
+
+/* Use inlined functions for supported systems */
+#if defined(_MSC_VER) && !defined(__cplusplus)   /* Visual Studio */
+  #define BLOSC_INLINE __inline  /* Visual C is not C99, but supports some kind of inline */
+#else
+  #if defined __STDC__ && defined __STDC_VERSION__ && __STDC_VERSION__ >= 199901L
```
kalvdans: Which could be written shorter as
```c
#elif __STDC_VERSION__ >= 199901L
```
Remember undefined macros evaluate to 0.

FrancescAlted: Good point.  I'll fix this after merging.

Issue #248, opened 2018-12-01
Bare inclusion of stdint.h in blosclz


alimanfoo:
https://github.com/Blosc/c-blosc/blob/ed0798f306de41179c2cd81696f241167a2d4a0f/blosc/blosclz.c#L17

This means c-blosc won't compile for Python extensions on PY27 on Windows.

jakirkham:
Added PR ( https://github.com/Blosc/c-blosc/pull/249 ) to fix it.

alimanfoo:
No worries at all, thanks for getting this in, look forward to 1.15.1
release.

PR #249, opened 2018-12-01
Use vendored `stdint.h` for older Visual Studio

jakirkham:
Fixes https://github.com/Blosc/c-blosc/issues/248

On older versions of Visual Studio, `stdint.h` is not included. In these cases, we need to use our vendored copy of `stdint.h` as a stand-in. For newer versions of Visual Studio or other compilers, `stdint.h` is generally included. So this workaround is unneeded in those cases and we handle them accordingly as well.

cc @alimanfoo

### DIFF
--- a/blosc/blosclz.c
+++ b/blosc/blosclz.c
@@ -14,8 +14,20 @@
 
 
 #include <stdio.h>
-#include <stdint.h>
 #include <stdlib.h>
+
+#if defined(_WIN32) && !defined(__MINGW32__)
+  #include <windows.h>
+  /* stdint.h only available in VS2010 (VC++ 16.0) and newer */
+  #if defined(_MSC_VER) && _MSC_VER < 1600
+    #include "win32/stdint-windows.h"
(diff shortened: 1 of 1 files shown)


esc:
FWIWhttps://github.com/Blosc/python-blosc/pull/166

 I can confirm that building 1.15.0 with python-blosc for 2.7 on windows works fine.

jakirkham:
Glad to hear it helps. I probably don’t deserve that much credit as I copy-pasted it from somewhere else in c-blosc. 😀

Thanks for the speedy review and merge. Also thanks for testing.

Agree a patch release would be nice if it’s not too much work.

FrancescAlted:
Release 1.15.1 is out.  Thanks!

esc:
@FrancescAlted super, thanks!!

alimanfoo:
Thanks @FrancescAlted!

jakirkham:
Thank you!

Issue #251, opened 2019-01-11
Thread deadlock bug in blosc_compress function


wenjuno:
There is a bug in `blosc_compress` function in `blosc/blosc.c` that can lead to thread deadlock. It's kind of obvious by just looking at the code so I haven't created a test case to reproduce the bug (I was bogged down by this bug while using hdf5-blosc).

Here is the relevant code snippet:
https://github.com/Blosc/c-blosc/blob/eb37eb815f75391b6f9663d481441458030262e8/blosc/blosc.c#L1380-L1395
The lock on the mutex will never be released when the code exits on line 1386 https://github.com/Blosc/c-blosc/blob/eb37eb815f75391b6f9663d481441458030262e8/blosc/blosc.c#L1386
or line 1389
https://github.com/Blosc/c-blosc/blob/eb37eb815f75391b6f9663d481441458030262e8/blosc/blosc.c#L1389
So the next call to `blosc_compress` will deadlock on trying to lock the mutex
https://github.com/Blosc/c-blosc/blob/eb37eb815f75391b6f9663d481441458030262e8/blosc/blosc.c#L1380

Right now, I have to use set environment variable `BLOSC_NOLOCK` to avoid this code block to workaround this issue.

I think a fix can be very simple. My idea would be to use `do-while(false)` block around codes between the locking and unlocking of mutex and use `break` to jump out of the block when there is an error, so mutex will always be unlocked, and at the end we check error code to determine whether to return `error` or `result`. But I'm open to other working solution.

Thanks.

FrancescAlted:
Another possibility would be to release the lock before the `return`, right?  Either way, I'd be happy to review a PR that fixes your use case.

wenjuno:
Thanks!

PR #252, opened 2019-01-12
Fix mutex lock leak in blosc_compress.

wenjuno:
Fix bug #251.


wenjuno:
Oops, should've realized that `result` and `error` have the same usage. I have removed the `error` and use the `result` exclusively for return code so I can avoid the added but unnecessary branch (checking `error`) at the end.

FrancescAlted:
Your PR looks good to me.  Thanks!

Issue #250, opened 2018-12-14
Assert bug in test_forksafe.c


jbms:
The first call to mu_assert in test_forksafe.c incorrectly reverses the order of the arguments, which means it doesn't have the intended effect.

PR #255, opened 2019-01-16
Fix bugs in test_forksafe.c

jbms:
The arguments to the mu_assert call were reversed, which resulted in
the test passing even though it was broken and not correctly testing
that blosc_decompress can be used after a call to `fork()`.

This change fixes the mu_assert call and fixes the bug in the test.


jbms:
This fixes issue #250.

FrancescAlted:
Oops, forgot about this one.  Thanks!

PR #256, opened 2019-01-16
Fix data race in shuffle.c host_implementation initialization.

### LINKED ISSUE #253: Data race in init_shuffle_implementation
There appears to be a data race in the init_shuffle_implementation function in shuffle.c (identified by ThreadSanitizer).

The problematic lines are:
```
  if (!implementation_initialized) {
    /* Initialize the implementation. */
    host_implementation = get_shuffle_implementation();

    /*  Set the flag indicating the implementation has been initialized. */
    implementation_initialized = 1;
  }
```
The issue is that the compiler is free to reorder the writes to host_implementation and implementation_initialized, and likewise could reorder the subsequent read from host_implementation to before the read from implementation_initialized, since no memory synchronization is specified.

The solution is to make implementation_initialized an atomic variable (e.g. using C11 _Atomic), or an equivalent compiler-specific builtin if you don't want to depend on C11.

jbms:
This fixes #253.


blosc/shuffle.c
```diff
@@ -13,6 +13,37 @@
...
+    init_routine();
+    InitOnceComplete(once_control, /*dwFlags=*/0, /*lpContext=*/NULL);
+  }
+  return 0;
+}
+
+#else
+#include <pthread.h>
+#endif
+
```
FrancescAlted: For consistency, I think it be worth adding `pthread_once()` to https://github.com/Blosc/c-blosc/blob/master/blosc/win32/pthread.c instead.  Would you agree?

FrancescAlted: Also, do you think this would be compatible with at least VS2008?  We need that for Python2 libraries that embed C-Blosc.

jbms: Adding it to pthread.h/pthread.c instead of inlining it into shuffle.c makes sense.  However, I am a bit confused by the fact that blosc.c #includes pthread.c, but in addition the CMakeLists.txt file sometimes includes win32/pthread.c in the SOURCES (which would lead to duplicate definitions).  I guess that condition (the system Threads library not being found) must never be satisfied in practice?  Wouldn't it make more sense to remove the #include pthread.c from blosc.c, and just rely on the build configuration to include win32/pthread.c ?

Regarding VS2008, I'm not sure what the interaction is between VS2008 and Windows API versions.  I don't have a way to test the Windows build other than Appveyor, and I see that the appveyor configuration for this package doesn't seem to test with VS2008.

jbms: Another issue that occurred to me is that there could be a conflict if on Windows the user links c-blosc, with its own partial pthread library that exports some pthread_* symbols, and another more full-featured pthread library that also exports some pthread_* symbols.  I guess this would only be an issue if static linking is used on Windows.  The solution to that problem would be to add to win32/pthread.h #define pthread_XXX blosc_internal_pthread_XXX  for each pthread symbol.

jbms:
I moved the pthread_once implementation to win32/pthread.h

I also added an appveyor configuration for VS2008.  That configuration seems to be failing, but for reasons that appear unrelated to the pthread_once code.

blosc/shuffle.c
FrancescAlted: > Adding it to pthread.h/pthread.c instead of inlining it into shuffle.c makes sense. However, I am a bit confused by the fact that blosc.c #includes pthread.c, but in addition the CMakeLists.txt file sometimes includes win32/pthread.c in the SOURCES (which would lead to duplicate definitions). I guess that condition (the system Threads library not being found) must never be satisfied in practice? Wouldn't it make more sense to remove the #include pthread.c from blosc.c, and just rely on the build configuration to include win32/pthread.c ?

Yes, AFAIK there is not official port of pthreads for Windows, so it is safe to assume that with win32/VS will never be double inclusion of the pthread.c and win32/pthread.c.  This has been working for years, and frankly, I would not touch this configuration.

> 
> Regarding VS2008, I'm not sure what the interaction is between VS2008 and Windows API versions. I don't have a way to test the Windows build other than Appveyor, and I see that the appveyor configuration for this package doesn't seem to test with VS2008.

That's correct, there is no provision for testing VS2008 in Appveyor.  This should be fixed somehow (honestly, I don't remember the reason why VS2008 is not tested anymore).

FrancescAlted: > Another issue that occurred to me is that there could be a conflict if on Windows the user links c-blosc, with its own partial pthread library that exports some pthread_* symbols, and another more full-featured pthread library that also exports some pthread_* symbols. I guess this would only be an issue if static linking is used on Windows. The solution to that problem would be to add to win32/pthread.h #define pthread_XXX blosc_internal_pthread_XXX for each pthread symbol.

As I said before, there is not a 'blessed' version of pthreads for Windows (mainly because Windows comes with its own threading library), so I don't think we should worry about potential duplicated symbols on the static Windows version of the library.

FrancescAlted:
Mmmhh, I triggered a new build and the appveyor still fails exactly in the same configuration than before, so probably we are facing to not just a random failure here.  FWIW, here it is the error:

```
13>Performing Post-Build Event...
75913>Error copying file (if different) from "C:/Users/appveyor/.conan/data/c-blosc/dev/francescalted/stable/build/6febb4e5bbf4f38ae4f73ee279a55264e9fb4bcf/build/blosc//blosc.dll" to "C:/Users/appveyor/.conan/data/c-blosc/dev/francescalted/stable/build/6febb4e5bbf4f38ae4f73ee279a55264e9fb4bcf/build/compat//blosc.dll".
76013>Project : error PRJ0019: A tool returned an error code from "Performing Post-Build Event..."
76113>Project : warning PRJ0018 : The following environment variables were not found:
76213>$(Configuration)
76313>Build log was saved at "file://c:\Users\appveyor\.conan\data\c-blosc\dev\francescalted\stable\build\6febb4e5bbf4f38ae4f73ee279a55264e9fb4bcf\build\compat\filegen.dir\Release\BuildLog.htm"
76413>filegen - 1 error(s), 0 warning(s)
```

Not sure what's provoking this tough.  Probably worth investigating more in deep.

jbms:
It turns out the problem was just that the cmake configuration needed to be fixed to work with VS2008.

FrancescAlted:
Excellent work @jbms .  Merging.

PR #257, opened 2019-01-16
Prefix all internal non-local symbols with "blosc_internal_"

### LINKED ISSUE #254: Prefixing symbol names
Many internal functions are given generic names like "shuffle".  They are given hidden visibility, which avoids linking problems when dynamically linking, but that doesn't help when using static linking.  To resolve this, all internal functions could be prefixed with "blosc_internal_".  If the maintainers are open to that change, I can provide a pull request.

jbms:
Also mark as static all functions that are only used within a single
translation unit.

This fixes #254.


FrancescAlted:
LGTM.  Thanks!

PR #258, opened 2019-01-28
Make blosc decompression safe for untrusted/possibly corrupted inputs

jbms:


jbms:
See the following benchmark results.

[results-lz4-suite-6-2097152-8-19-safe.txt](https://github.com/Blosc/c-blosc/files/2804192/results-lz4-suite-6-2097152-8-19-safe.txt)
results-lz4-suite-6-2097152-8-19-safe
[results-lz4-suite-6-2097152-8-19-fast.txt](https://github.com/Blosc/c-blosc/files/2804193/results-lz4-suite-6-2097152-8-19-fast.txt)
results-lz4-suite-6-2097152-8-19-fast
[results-blosclz-suite-6-2097152-8-19-safe.txt](https://github.com/Blosc/c-blosc/files/2804194/results-blosclz-suite-6-2097152-8-19-safe.txt)
results-blosclz-suite-6-2097152-8-19-safe
[results-blosclz-suite-6-2097152-8-19-fast.txt](https://github.com/Blosc/c-blosc/files/2804195/results-blosclz-suite-6-2097152-8-19-fast.txt)
results-blosclz-suite-6-2097152-8-19-fast

bench/plot-speeds.py
```diff
@@ -27,13 +27,18 @@ def get_values(filename):
     for line in f:
         if line.startswith('-->'):
             tmp = line.split('-->')[1]
-            nthreads, size, elsize, sbits, codec, shuffle = [i for i in tmp.split(', ')]
+            parts = tmp.split(', ')
+            nthreads, size, elsize, sbits, codec, shuffle = parts[:6]
+            safe = 'unsafe'
```
FrancescAlted: For consistency, here it should be 'fast', I think.

jbms: Fixed.

FrancescAlted:
Yes, it looks like there is almost no impact in performance.  Here are my own results using an Intel processor with 4 physical cores (Xeon E3-1245 v5 @ 3.50GHz).

For blosclz:

image

image

For lz4:

image

image

So I am mostly sold on this PR 👍

FrancescAlted:
Two points before going ahead with this:

1. In the code there is a mix of 'unsafe' (mainly in the API) and 'fast' (mainly in the bench).  We should be consistent here.  I slightly prefer the 'fast' wording (even if benchmarks say that there is almost no difference in performance, keeping '_unsafe' versions may look confusing for the user).  But still, it is important to stress out in the docstrings that 'fast' also means 'unsafe'.

2. Are the 'safe' versions in this PR (i.e. the defaults) passing the [libfuzzer](https://llvm.org/docs/LibFuzzer.html) checks?  I suppose this is not the case, but I am curious if this check can be easily added to the current CI configuration (even if it is only for Linux).

jbms:
I think _unsafe may be better than _fast because when reviewing code _unsafe immediately indicates to the reader that should check the usage while _fast might just indicate that it is an optimized implementation and is less likely to receive scrutiny (on the other hand, LZ4 uses the "fast" term), but if you would like me to change it to "fast" I can do that.

Regarding the libfuzzer checks, yes I haven't discovered any problems with the safe versions of decompress and getitem, however I haven't run the fuzzer all that long, maybe about an hour.  The way fuzzing works is that there is no set length of time to run it for --- the longer it runs, the more likely it is to discover new code paths and new problems.  For this reason the actual fuzzing process doesn't fit all that well into the Travis/Appveyor model.  However, libfuzzer does generate a "corpus" of example inputs that cover all of the code paths that have been discovered --- and it would make sense to have a test that feeds through these examples to verify there aren't any regressions.  The separate issue is that the fuzzing was done through an internal build system we use at Google, and it may take a bit of work to set it up with cmake.

jbms:
If you'd like, this project could be added to oss-fuzz (https://github.com/google/oss-fuzz) to receive continuous fuzz testing.

FrancescAlted:
> I think _unsafe may be better than _fast because when reviewing code _unsafe immediately indicates to the reader that should check the usage while _fast might just indicate that it is an optimized implementation and is less likely to receive scrutiny (on the other hand, LZ4 uses the "fast" term), but if you would like me to change it to "fast" I can do that.

Well, I am wondering now that, provided how close in performance are the _safe (default) and _unsafe versions, if it might be better to just keep the defaults and remove the _unsafe.  The only thing that I am a bit afraid is if the safe versions would reduce performance more in a real data scenario than in the synthetic data tested here.  What's your feeling about this?

> 
> Regarding the libfuzzer checks, yes I haven't discovered any problems with the safe versions of decompress and getitem, however I haven't run the fuzzer all that long, maybe about an hour. The way fuzzing works is that there is no set length of time to run it for --- the longer it runs, the more likely it is to discover new code paths and new problems. For this reason the actual fuzzing process doesn't fit all that well into the Travis/Appveyor model. However, libfuzzer does generate a "corpus" of example inputs that cover all of the code paths that have been discovered --- and it would make sense to have a test that feeds through these examples to verify there aren't any regressions. The separate issue is that the fuzzing was done through an internal build system we use at Google, and it may take a bit of work to set it up with cmake.

I see.  At any rate, it is reasssuring that this PR is fuzzer clean, even if it ran for a relatively short time.

FrancescAlted:
> If you'd like, this project could be added to oss-fuzz (https://github.com/google/oss-fuzz) to receive continuous fuzz testing.

That would be great indeed.  Would you like to act as a proxy for the Blosc project and ask for including it?  As a contact person you can put my name.

bench/bench.c
```diff
@@ -391,7 +393,7 @@ int main(int argc, char *argv[]) {
   strncpy(usage, "Usage: bench [blosclz | lz4 | lz4hc | snappy | zlib | zstd] "
           "[noshuffle | shuffle | bitshuffle] "
           "[single | suite | hardsuite | extremesuite | debugsuite] "
-          "[nthreads] [bufsize(bytes)] [typesize] [sbits]", 255);
+          "[nthreads] [bufsize(bytes)] [typesize] [sbits] [fast | safe]", 255);
```
FrancescAlted: For consistency, rename this to '[safe | unsafe]'

bench/bench.c
```diff
@@ -489,7 +491,18 @@ int main(int argc, char *argv[]) {
     rshift = atoi(argv[7]);
   }
 
-  if ((argc >= 9) || !(single || suite || hard_suite || extreme_suite)) {
+  if (argc >= 9) {
+    if (strcmp(argv[8], "fast") == 0) {
```
FrancescAlted: Ditto

bench/plot-speeds.py
```diff
@@ -27,13 +27,18 @@ def get_values(filename):
     for line in f:
         if line.startswith('-->'):
             tmp = line.split('-->')[1]
-            nthreads, size, elsize, sbits, codec, shuffle = [i for i in tmp.split(', ')]
+            parts = tmp.split(', ')
+            nthreads, size, elsize, sbits, codec, shuffle = parts[:6]
+            safe = 'fast'
```
FrancescAlted: Ditto

bench/bench.c
```diff
@@ -215,7 +215,7 @@ void do_bench(char *compressor, char *shuffle, int nthreads, int size, int elsiz
      if (retcode) abort();
   }
 
-  fprintf(ofile, "--> %d, %d, %d, %d, %s, %s\n", nthreads, size, elsize, rshift, compressor, shuffle);
+  fprintf(ofile, "--> %d, %d, %d, %d, %s, %s, %s\n", nthreads, size, elsize, rshift, compressor, shuffle, unsafe ? "fast" : "safe");
```
FrancescAlted: Ditto

FrancescAlted:
One last consideration.  In the future I'd like to add support for Intel's accelerated version of LZ4, and after having a look at:

https://software.intel.com/en-us/ipp-dev-reference-lz4-compression-functions

it looks strange to me that Intel seems to have produced just a safe version of the 'encode' functions and not for the 'decode' ones (contrarily to this PR).  Do you have experience with Intel's IPP?  Maybe it is worth asking them.

FrancescAlted:
Just to provide a hint on the increase of performance that one can get by using the accelerated LZ4 codec in Intel's IPP, here it is a comparison quickly made on my MacBook (Core-i5 @2.9 GHz, 2 cores + hyper-threading).

Plain LZ4 (1.8.2):

lz4-plain

Accelerated LZ4 in Intel's IPP (2019.1.144):

lz4-ipp

so the boost can be quite impressive (up to 2x for single-threaded and lowest compression level).

bench/bench.c
jbms: Done.

bench/bench.c
jbms: Done.

bench/plot-speeds.py
jbms: Done.

bench/bench.c
jbms: Done.

jbms:
I don't have experience with IPP.  I think the decode functions may always be safe, since providing only an unsafe version would be quite unwise, and there is an ippStsMemAllocErr "Indicates an error if the size of the allocated memory is not sufficient for decompression", although I suppose it is possible that refers to the dictionary.  It is unfortunate that the documentation is not more clear.   I suspect that the caller is supposed to initialize *pDstLen with the size of the destination buffer.

I can't figure out what "safe" means in the context of the encoding functions.  The difference in signature seems to be that the "safe" functions take a int* pSrcLen parameter, while the regular functions take a int srcLen parameter.  The documentation doesn't seem to indicate what purpose the pSrcLen pointer serves.  It would probably make sense to ask Intel for clarification.

FrancescAlted:
FYI I have asked Intel about LZ4/IPP safety: https://software.intel.com/node/804325

PR #262, opened 2019-02-06
Add blosc_internal prefix to _xgetbv and __cpuid definitions in shuffle.c

jbms:
This fixes the build for clang with -march=haswell, where including
immintrin.h includes xsaveintrin.h, which defines _xgetbv.  This
change additional renames __cpuid to prevent possible future breakages
and for conformance with the ISO C standard which reserves all global
names beginning with an underscore for use by the compiler.

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -103,11 +103,12 @@ static blosc_cpu_features blosc_get_cpu_features(void) {
 /*  _xgetbv is only supported by VS2010 SP1 and newer versions of VS. */
 #if _MSC_FULL_VER >= 160040219
   #include <immintrin.h>  /* Needed for _xgetbv */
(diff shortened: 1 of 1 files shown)


FrancescAlted:
In this commit I am afraid that I am only able to spot the renaming, but not the issue with clang and xsaveintrin.h.  Could you shed some light on where is this addressed?

FrancescAlted:
Ah, I think you meant that the rename of `xgetbv` is the actual fix.  Yep, if you confirm I will merge this.

jbms:
Yes, that's right --- some versions of clang define _xgetbv which creates a conflict.  Trying to detect when it is defined would be tricky and likely to miss some cases, and renaming is simple.

PR #272, opened 2019-07-28
Install .dll in prefix/bin

jschueller:


FrancescAlted:
Ok, I suppose this is valid only for Windows machines right?  Could you provide a patch that installs .DDL in `prefix/bin`  only on Windows?

jschueller:
This applies for all win32 targets, ie not only on windows hosts, but also when cross-compiling from Linux to windows (eg mingw).

FrancescAlted:
Yes, but my point is that your patch also affects to Unix machines, and probably this is not desired there, right?

jschueller:
No, this does not affect unix machines, why would you assume that ?

jschueller:
No, this does not affect unix machines, shared libraries are treated via the LIBRARY component (so installed into prefix/lib). On win32 this is different, dll is RUNTIME, and the import library is treated via ARCHIVE.

FrancescAlted:
Ah, that makes more sense now.  Let me merge and see what's is happening on my Linux box.

jschueller:
Not much will happen :)

FrancescAlted:
Yup, seems to work as intended.  Thanks!

PR #276, opened 2019-09-05
WIP: allow threading backend to be replaced by caller

stevengj:
This PR adds a new function `blosc_set_threads_callback` that allows the caller to register a callback in order to change the threading backend at runtime — instead of spawning its own threads and passing work to them with a mutex, Blosc will pass an array of work (an array of `thread_context*`)  to the callback, which can execute the work in serial or in parallel using whatever mechanism it wants.

Motivation: this allows Blosc threading to be *composable* with caller multithreading, instead of having Blosc and the caller fight over the same CPUs.    In particular, this enables Blosc threading to be composable with:

* Julia's [new partr threading scheduler](https://julialang.org/blog/2019/07/multithreading).
* Callers using [Intels's Threading Building Blocks (TBB)](https://en.wikipedia.org/wiki/Threading_Building_Blocks).
* Callers using Cilk, OpenMP, etcetera.

The design and implementation is very similar to the new pluggable threading backend for FFTW (FFTW/fftw3#175), which worked out very well for us.   Aside from a trivial refactoring of the `t_blosc` function, required changes seem pretty minimal — only a couple of dozen lines, with a practical overhead of a single `if` statement in the `parallel_blosc` function for people not using this feature.

Marking it as a WIP since I haven't tested it yet other than checking that it compiles, but I wanted to get some early feedback.

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -1731,10 +1731,9 @@ int blosc_getitem_unsafe(const void *src, int start, int nitems, void *dest) {
   return blosc_getitem_impl(src, start, nitems, dest, /*unsafe=*/1);
 }
 
-/* Decompress & unshuffle several blocks in a single thread */
-static void *t_blosc(void *ctxt)
+/* execute single compression/decompression job for a single thread_context */
+static void t_blosc_do_job(struct thread_context *context)
 {
-  struct thread_context* context = (struct thread_context*)ctxt;
   int32_t cbytes, ntdest;
   int32_t tblocks;              /* number of blocks per thread */
   int32_t leftover2;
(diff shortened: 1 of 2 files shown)


blosc/blosc.c
```diff
@@ -1731,10 +1747,11 @@ int blosc_getitem_unsafe(const void *src, int start, int nitems, void *dest) {
   return blosc_getitem_impl(src, start, nitems, dest, /*unsafe=*/1);
 }
 
-/* Decompress & unshuffle several blocks in a single thread */
-static void *t_blosc(void *ctxt)
+/* execute single compression/decompression job for a single thread_context */
+static void t_blosc_do_job(void *ctxt)
```
stevengj: Note that this code is a trivial refactoring — the `t_blosc_do_job` is simply cut-and-pasted out of the original `t_blosc` function.

stevengj:
I'll move this to c-blosc2, as [suggested](https://github.com/Blosc/c-blosc/issues/277#issuecomment-528735057) by @FrancescAlted.  Looks like an almost identical patch will work there.

stevengj:
Closing in favor of https://github.com/Blosc/c-blosc2/pull/81

PR #280, opened 2019-09-26
Add bounds check back to beginning of blosclz_decompress

jbms:
The bounds check before the initial read was lost in
dfcd0e949a89bf4bdaab71fb0805625f7f7a284a.

I haven't verified that this is exploitable, but it seems wise to have
it there in any case.


FrancescAlted:
Frankly, I don't see how this can be exploitable, but the suggestion seems safe to me, so merging.

PR #281, opened 2019-10-21
Add config export to cmake build for downstream find_package

tamasmeszaros:
Part of integration into [PrusaSlicer](https://github.com/prusa3d/PrusaSlicer) I wrote some cmake logic to export targets for downstream users. CMake variables are changed to imported targets everywhere in the build script `target_link_libraries` calls to support the newer cmake paradigm.


tamasmeszaros:
I think the CI hits the timeout.

CMakeLists.txt
```diff
@@ -71,7 +71,7 @@
 #    DEV: static includes blosc.a and blosc.h
 
 
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.1) # Threads::Threads target available from 3.1
```
FrancescAlted: This is quite a strong requirement for a library to tries to be compatible with pretty old OS distros.  I suggest to use the [c-blosc2](https://github.com/Blosc/c-blosc2) project for such a innovative patches.  Thanks!

tamasmeszaros: @FrancescAlted Thank you for the feedback, I completely understand. I will close this PR in this case, it already got too heavy.

The requirement might be lowered if not for the Threads target but I don't think its worth the hassle.

PR #292, opened 2020-05-02
Minor cleanup to GitHub Actions CI

nmoinvaz:


FrancescAlted:
Thanks.  The `build.py` belongs to Conan package system, and it has been deprecated (wheels are probably the best bet for users, and the only binaries that we would like to support), so probably all the Conan related files should be removed.  I'll create a new ticket for this.

FrancescAlted:
Conan support has been deprecated in commit ed7ad70.

Issue #285, opened 2020-02-20
Blosc should use CXX linker when shared and snappy static


SpaceIm:
I'm creating a conan recipe of Blosc for https://github.com/conan-io/conan-center-index

Except msvc, I had lot of undefined reference to libstdc++ symbols (due to Snappy), when linking Blosc into an executable.

Seems like Blosc doesn't use C++ linker if external Snappy (static) and shared Blosc.

I've solved this pb by patching blosc/CMakeLists.txt:
````patch
@@ -193,6 +135,9 @@ endif()
 
 if (BUILD_SHARED)
     target_link_libraries(blosc_shared ${LIBS})
+    if(NOT DEACTIVATE_SNAPPY)
+        set_target_properties(blosc_shared PROPERTIES LINKER_LANGUAGE CXX)
+    endif()
     target_include_directories(blosc_shared PUBLIC ${BLOSC_INCLUDE_DIRS})
 endif()
 
````

(you can also have a look here: https://github.com/SpaceIm/conan-c-blosc)

If both Snappy and Blosc are static, responsibility of using C++ linker lies to the consumer of Blosc...

FrancescAlted:
Closing as per #295.

Issue #201, opened 2017-08-24
Disable Snappy support by default


FrancescAlted:
Snappy is a C++ codec that creates unnecessary complexity in building tools (see https://github.com/Blosc/python-blosc/issues/127#issuecomment-324706140), so provided that there are better codecs giving better speed and compression (e.g. LZ4), I propose it to be disabled by default.

nmoinvaz:
This looks like it is done, do you want to go a step further and remove building snappy from sources but allow testing against external snappy (similar to c-blosc2)?

FrancescAlted:
Yes, that would be cool.

PR #295, opened 2020-05-09
Only allow external snappy support

nmoinvaz:
See discussion from #201. 

Also once this is merged #285 can be closed.


FrancescAlted:
LGTM.  Thanks!

PR #299, opened 2020-06-05
propagate pthread_create errors back to the user

hotpxl:
Sometimes `pthread_create` would fail, and we should propagate and return the error.
Otherwise in `parallel_blosc` it would get stuck on
```
/* Synchronization point for all threads (wait for initialization) */
WAIT_INIT(-1, context);

/* Synchronization point for all threads (wait for finalization) */
WAIT_FINISH(-1, context);
```
waiting forever.


FrancescAlted:
Good point.  Thanks!

PR #300, opened 2020-06-10
Redefine blosc_internal_xgetbv for platforms with immintrin.h

mgorny:
Commit 2f2d876 renamed _xgetbv to blosc_internal_xgetbv.  Commit 0b8608b
made it conditional to immintrin.h not being included, apparently
in order to prevent _xgetbv collision.  However, since the block
no longer defined _xgetbv but blosc_internal_xgetbv, it means that
the latter is not defined if the condition is false.  #define it to
_xgetbv to fix missing symbol.


FrancescAlted:
Thanks!

PR #301, opened 2020-06-17
Add fuzzers to c-blosc project

nmoinvaz:
* Adds simple fuzzers for testing `blosc_compress` and `blosc_decompress`.
  * Fuzzers are linked against fuzzing library for use with OSS-Fuzz
* Adds cmake tests using the fuzzers and tests against all the files in `compat` folder, all those files are a good seed corpus.
* Adds `BUILD_FUZZERS` cmake option for enabling the fuzzers and their tests.

I used 2 space indent because that seems to be what you used in blosc code.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -143,7 +143,7 @@ jobs:
       run: |
         mkdir ${{ matrix.build-dir || '.not-used' }}
         cd ${{ matrix.build-dir || '.' }}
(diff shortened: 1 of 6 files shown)


nmoinvaz:
I added `-DBUILD_FUZZERS=ON` to GitHub Actions. 

It should be ready for merge now. I have been running `compress_fuzzer` for 8 hours now with no issues. After this is merged, I can then update `c-blosc` project in oss-fuzz and let Google's machines see if they can find anything else.

PR #302, opened 2020-06-18
Check that source size is greater than BLOSC_MIN_HEADER_LENGTH

nmoinvaz:
This is so that `blosc_decompress` does not read past end of source buffer if source size is `BLOSC_MIN_HEADER_LENGTH`. 

It is unclear from [README_HEADER](https://github.com/Blosc/c-blosc/blob/master/README_HEADER.rst) whether or not `nbytes` or `ctbytes` include the size of the header.

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -1450,7 +1450,7 @@ static int blosc_run_decompression_with_context(struct blosc_context* context,
   context->compressedsize = sw32_(context->src + 12); /* compressed buffer size */
   context->bstarts = (uint8_t*)(context->src + 16);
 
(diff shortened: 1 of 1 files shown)


nmoinvaz:
I'm not sure, I think this may be a bug in my fuzzer.

FrancescAlted:
nbytes does not include the header, whereas ctbytes (now cbytes in README_HEADER) does. I have updated the doc accordingly.

nmoinvaz:
Thanks! It might be good idea to update it in blosc2 too because I looked there to see if it had that info and I didn’t see it there either.

PR #304, opened 2020-06-22
Group third-party source files with CMake

nmoinvaz:
This PR groups the source files into folders with the name of the third-party library. 

At least in Visual Studio this looks like:

image

I can make a similar PR to C-blosc2.


FrancescAlted:
That's pretty cool!

PR #309, opened 2020-07-27
BUILD_FUZZERS require BUILD_STATIC

mgorny:
Fail with an explicit error if BUILD_FUZZERS is enabled without
BUILD_STATIC, and default the former to the value of the latter.
This resolves ugly CMake warnings and build failure when static
libraries are disabled but fuzzers default to be enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ option(BUILD_SHARED
 option(BUILD_TESTS
     "Build test programs from the blosc compression library" ON)
 option(BUILD_FUZZERS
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Thanks @mgorny !

nmoinvaz:
This will probably need to be cherry-picked to Blosc-2 as well.

FrancescAlted:
> This will probably need to be cherry-picked to Blosc-2 as well.

Done in https://github.com/Blosc/c-blosc2/commit/7bfcc322c086f41df8188a493c989af21e3707d2

PR #318, opened 2021-06-16
CMake: Fix PThread Flag

ax3l:
## Addressed Issue

On recent manylinux builds for ppc64le, I realized problems with
```
undefined reference to `pthread_atfork'
```

at linktime when propagating a static blosc library to downstream libraries (all via CMake).

## Fix: CMake Control Variable

An outdated CMake flag that would request the right pthread flags is used:

Old:
`CMAKE_THREAD_PREFER_PTHREAD` ([docs](https://cmake.org/cmake/help/v3.2/module/FindThreads.html))

New:
`THREADS_PREFER_PTHREAD_FLAG` ([docs](https://cmake.org/cmake/help/latest/module/FindThreads.html))

## Fix: CMake Thread Target

Prefer `Threads::Threads` for CMake 3.1+: This target does potentially contain CXX flags and linker flags and should thus be preferred.

It should also address the already work-arounded pthread linker issue in one of the tests that @jack-pappas added in #123

## To Do

- [x] finish tests

### DIFF
--- a/blosc/CMakeLists.txt
+++ b/blosc/CMakeLists.txt
@@ -74,11 +74,19 @@ if(WIN32)
         message(STATUS "using the internal pthread library for win32 systems.")
         set(SOURCES ${SOURCES} win32/pthread.c)
     else(NOT Threads_FOUND)
-        set(LIBS ${LIBS} ${CMAKE_THREAD_LIBS_INIT})
+        if(CMAKE_VERSION VERSION_LESS 3.1)
+            set(LIBS ${LIBS} ${CMAKE_THREAD_LIBS_INIT})
+        else()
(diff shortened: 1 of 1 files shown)


ax3l:
We should maybe also use the new `Threads::Threads` target (CMake 3.1) instead of the result variables?

kalvdans (approved):

ax3l:
I think at least for the static build, we also need to change
```cmake
target_link_libraries(blosc_static ${LIBS})
```
to
```cmake
target_link_libraries(blosc_static PUBLIC ${LIBS})
```

to propagate lib flags such as `-pthread`

FrancescAlted:
That looks good to me.  Thanks for the fix!

PR #321, opened 2021-09-01
Add public function blosc_compression_bound

### LINKED ISSUE #320: Is there a function like compress_bound() ?
When I use zlib, I tend to reserve just enough size in my buffers with `::compressBound`. LZ4 also provides this feature, and I assume all compression libs do. It would be very useful if blosc provided a function for this

Guekka:
This function can be used to determine the output buffer length when
compressing.

Fixes #320

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -1422,6 +1422,52 @@ int blosc_compress(int clevel, int doshuffle, size_t typesize, size_t nbytes,
   return result;
 }
 
+/* The public api for buffer size */
+size_t blosc_compression_bound(size_t cursize)
+{
+  switch (g_compressor)
+  {
+  case BLOSC_BLOSCLZ:
+    // from doc : The output buffer must be at least 5% larger than the input buffer
+    // and can not be smaller than 66 bytes.
+    return max(66, (size_t)ceil(1.05 * cursize));
+
+#if defined(HAVE_LZ4)
+  case BLOSC_LZ4:
+  case BLOSC_LZ4HC:
+    return (size_t)LZ4_compressBound((int)cursize);
+#endif /* HAVE_LZ4 */
+
+#if defined(HAVE_SNAPPY)
+  case BLOSC_SNAPPY:
+    return snappy_max_compressed_length(cursize);
+#endif /* HAVE_SNAPPY */
+
+#if defined(HAVE_ZLIB)
+  case BLOSC_ZLIB:
+    return (size_t)compressBound((uLong)cursize);
+#endif /* HAVE_ZLIB */
+
+#if defined(HAVE_ZSTD)
+  case BLOSC_ZSTD:
+    return ZSTD_compressBound(cursize);
+#endif /* HAVE_ZSTD */
+
+  default:
+  {
+    const char *compname;
+    compname = clibcode_to_clibname(g_compressor);
+    if (compname == NULL) {
+        compname = "(null)";
+    }
+    fprintf(stderr, "Blosc has not been compiled with '%s' ", compname);
+    fprintf(stderr, "compression support.  Please use one having it.");
+    return -5;    /* signals no compression support */
+    break;
+  }
+  }
+}
+
 static int blosc_run_decompression_with_context(struct blosc_context* context,
                                                 const void* src,
                                                 void* dest,
--- a/blosc/blosc.h
+++ b/blosc/blosc.h
@@ -224,6 +224,11 @@ BLOSC_EXPORT int blosc_compress(int clevel, int doshuffle, size_t typesize,
 				size_t nbytes, const void *src, void *dest,
 				size_t destsize);
 
+/**
+  Returns the maximal size of the compressed representation of input data
+  that is "cursize" bytes in length.
+  */
+BLOSC_EXPORT size_t blosc_compression_bound(size_t cursize);
 
 /**
   Context interface to blosc compression. This does not require a call


blosc/blosc.c
```diff
@@ -1422,6 +1422,52 @@ int blosc_compress(int clevel, int doshuffle, size_t typesize, size_t nbytes,
...
+  default:
+  {
+    const char *compname;
+    compname = clibcode_to_clibname(g_compressor);
+    if (compname == NULL) {
+        compname = "(null)";
+    }
+    fprintf(stderr, "Blosc has not been compiled with '%s' ", compname);
+    fprintf(stderr, "compression support.  Please use one having it.");
+    return -5;    /* signals no compression support */
```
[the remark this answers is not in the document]
Guekka: Could we instead return 0 to signal failure?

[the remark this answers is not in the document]
kalvdans: Yes, I think that is better as it will never return 0 on success.

kalvdans:
Can you add the same function for compression contexts? (i.e. the threadsafe API without global variables)

blosc/blosc.h
```diff
@@ -224,6 +224,13 @@ BLOSC_EXPORT int blosc_compress(int clevel, int doshuffle, size_t typesize,
 				size_t nbytes, const void *src, void *dest,
 				size_t destsize);
 
+/**
+  Returns the maximal size of the compressed representation of input data
+  that is "cursize" bytes in length.
+
+  Returns 0 on failure.
```
kalvdans, line 231:
```suggestion
  Returns 0 and prints an error to stderr on failure.
```

Guekka:
> Can you add the same function for compression contexts? (i.e. the threadsafe API without global variables)

Sure. How would you like it?
- duplicate code
- `blosc_compression_bound` calls `blosc_compression_bound_context` (involves `blosc_context` construction)
- private helper function used by both

kalvdans:
> > Can you add the same function for compression contexts? (i.e. the threadsafe API without global variables)
> 
> Sure. How would you like it?
> 
>     * duplicate code
> 
>     * `blosc_compression_bound` calls `blosc_compression_bound_context` (involves `blosc_context` construction)
> 
>     * private helper function used by both

I'm not the maintainer but I think a private helper function is cleanest. Thanks for discussing this before coding!

blosc/blosc.c
```diff
@@ -1422,6 +1422,63 @@ int blosc_compress(int clevel, int doshuffle, size_t typesize, size_t nbytes,
   return result;
 }
 
+size_t blosc_compression_bound_helper(size_t cursize, int compressor)
```
kalvdans, line 1425:
```suggestion
static size_t blosc_compression_bound_helper(size_t cursize, int compressor)
```

blosc/blosc.c
```diff
@@ -1422,6 +1422,63 @@ int blosc_compress(int clevel, int doshuffle, size_t typesize, size_t nbytes,
...
+    }
+    fprintf(stderr, "Blosc has not been compiled with '%s' ", compname);
+    fprintf(stderr, "compression support.  Please use one having it.");
+    return 0;    /* failure */
+    break;
+  }
+  }
+}
+
+size_t blosc_compression_bound_context(size_t cursize, struct blosc_context* context)
```
kalvdans, line 1470: Please include this function in the public header file as well, and duplicate the documentation from `blosc_compression_bound()`. After that, at least I'm happy with the changes and we will have to wait for someone with commit rights :)

Guekka: Great, thank you for your responsiveness

Guekka: Actually, I don't think I can. That struct is private

Guekka:
There's something wrong. Blosc consistently uses more space than `blosc_compression_bound` (with `zlib`)

Could it be that `blosc` has some overhead? Gonna investigate

Guekka:
>Small overhead on non-compressible data: only a maximum of (32 + 4 * nblocks_used) additional bytes over the source buffer length are needed to compress every input.

From https://www.blosc.org/pages/blosc-in-depth/

So `compress_bound` would actually need to know the number of blocks used. I will think about it. If the maintainer passes by, I'd be interested in some input

Guekka:
This commit seems to fix compression errors. But I do not know Blosc enough to determine if it will always work

blosc/blosc.c
kalvdans: You are right. Looking at other _ctx functions, it seems all necessary information is passed as arguments instead of having a context pointer. In this case it would be (I don't know if blocksize is needed):

  `compressor`: the string representing the type of compressor to use.
  `blocksize`: the requested size of the compressed blocks.  If 0, an
   automatic blocksize will be used.

FrancescAlted:
> > Small overhead on non-compressible data: only a maximum of (32 + 4 * nblocks_used) additional bytes over the source buffer length are needed to compress every input.
> 
> From https://www.blosc.org/pages/blosc-in-depth/
> 
> So `compress_bound` would actually need to know the number of blocks used. I will think about it. If the maintainer passes by, I'd be interested in some input

There was an error there, sorry.  The actual overhead is 32 bytes for Blosc2 and 16 bytes for Blosc1.  This has been amended in the website.

blosc/blosc.c
FrancescAlted: In general, I'd say that the function can be much simplified using just:

```
static size_t blosc_compression_bound_helper(size_t cursize) {
 return BLOSC_MAX_OVERHEAD + cursize;
}
```

I don't know if having such a function buys a lot for the user though.  Although if it mimics existing function in zlib, maybe it saves some time for users looking for this.

Guekka: Do you think so? Each underlying compressor requires a different amount of size

FrancescAlted: Right, but there is an internal check controlling the final size of the compressed buffer.  If it is larger than `BLOSC_MAX_OVERHEAD + nbytes`, the compression stops and a literal copy is returned instead (plus the header).

Guekka: I see. Then this function loses much usefulness. A simple documentation change would be enough

FrancescAlted: I'd think that the existing doc in https://github.com/Blosc/c-blosc/blob/master/blosc/blosc.h#L163 should be enough (but feel free to open a new PR for amending this differently).

FrancescAlted:
After some discussion, I think we can close this.

PR #328, opened 2021-12-06
CMake: Use GNUInstallDirs when installing files.

### LINKED ISSUE #190: pkgconfig file is installed in wrong dir
https://github.com/Blosc/c-blosc/commit/499d52af31944d05593ab3abc21a4f8728fe5712 hardcodes "/lib/", but it should use something like ${LIBDIR} (`/usr/lib64`, `/usr/lib`, `/usr/<triple>/lib`, etc.). I tried to make a patch, but cmake apparently does not define a standard variable for this esoteric concept, so I gave up.

havardAasen:
This pr is almost identical to the two commits in c-blosc2 Blosc/c-blosc2@aef01bb4 and Blosc/c-blosc2@d53783ce785318f4ca98c40880226c8ba71285c1, and will provide the same benefits.

This will also fix #190

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,7 +122,7 @@ option(PREFER_EXTERNAL_ZSTD
     "Find and use external Zstd library instead of included sources." OFF)
 
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
-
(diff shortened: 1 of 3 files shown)


FrancescAlted:
LGTM.  Thanks!

PR #329, opened 2022-02-08
Allow to expose symbols

oshogbo:
We would like to propose a new option to CMake, which allows exposing functions to the linker.

Some project like https://github.com/google/sandboxed-api requires this to build a sandbox.

By default, the option will still be disabled, but it would be good to allow the user to change this setting.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,8 @@
 #       build fuzz test programs and generates the "test" target
 #   BUILD_BENCHMARKS: default ON
 #       build the benchmark program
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Is that needed for some open source project?  Just curious ;-)

oshogbo:
I'm not sure I follow.
The sanboxed-api is an open source project.
We trying to get traction with it.

FrancescAlted:
Ah, it is just for sandbox-api then.  Ok, thanks!

PR #338, opened 2022-07-15
dos2unix

DimitriPapadopoulos:
Change the only file with Windows CRLF line terminators to Unix LF line terminators.


FrancescAlted:
Looks good.

Issue #334, opened 2022-03-21
macOS: Universal2 build fails on SSE2


ax3l:
Hi,

We ship openPMD-api as binary wheels for python users containing c-blosc  for postprocessing tasks.

With ARM being strong with the Apple M1 processors, we want to build `universal2` wheels that contain both x86_64 and arm64 code. We produce them on macOS 11.0 x86_64 GitHub actions runners.

Multi-arch builds for macOS are very easy with CMake, we just need to set:
```bash
export CMAKE_OSX_ARCHITECTURES="arm64;x86_64"
```

Unfortunately, `c-blosc` SSE2 breaks the build for SSE2. The only alternative I have is to disable SSE2 acceleration altogether at the moment, which means that `x86_64` code in the universal binary will not be accelerated either. Is there maybe a way to make this more flexible? :)

Refs.: https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary

X-ref: https://github.com/openPMD/openPMD-api/pull/1233

t20100:
Hi,

There's a similar issue when building [`hdf5plugin`](https://github.com/silx-kit/hdf5plugin) for `universal2` Python (see https://github.com/silx-kit/hdf5plugin/issues/186) which include blosc as a HDF5 compression filter.

ax3l:
The latest XCode LLVM that Apple ships aka `AppleClang` in CMake.

kalvdans:
Thanks, I can't find a list of predefined macros for AppleClang but if you run the compiler like `AppleClang --target=... -mcpu=... -E - -dM </dev/null` it will output all the pre-defined preprocessor macros (similar works for gcc, too)

Build with `make VERBOSE=1` to see the actual compiler binary and correct arch flags it uses.

PR #347, opened 2022-10-20
Enable macos `universal2` build with AVX2/SSE2 optimisation for x86_64 target

t20100:
This PR proposes a fix to allow building `c-blosc` as part of a Python package that is built as a `universal2` binary package on macos with enabled AVX2/SSE2 optimisation for the x86_64 target only.

To achieve this, `c-blosc` should support defining `SHUFFLE_AVX2_ENABLED` and `SHUFFLE_SSE2_ENABLED` and be able to build all files (including `*-avx2.c` and `*-sse2.c`) even when AVX2/SSE2 are not supported.

This is done by toggling AVX2/SSE2 with both `SHUFFLE_AVX2_ENABLED`, `SHUFFLE_SSE2_ENABLED` and `__AVX2__`, `__SSE2__`.

Related to #334
Related to https://github.com/silx-kit/hdf5plugin/issues/186

### DIFF
--- a/blosc/bitshuffle-avx2.c
+++ b/blosc/bitshuffle-avx2.c
@@ -19,8 +19,8 @@
 
 /* Make sure AVX2 is available for the compilation target and compiler. */
 #if !defined(__AVX2__)
-  #error AVX2 is not supported by the target architecture/platform and/or this compiler.
-#endif
+  #warning AVX2 is not supported by the target architecture/platform and/or this compiler.
+#else
 
 #include <immintrin.h>
 
@@ -243,3 +243,5 @@ int64_t blosc_internal_bshuf_untrans_bit_elem_avx2(void* in, void* out, const si
 
     return count;
 }
+
+#endif /* !defined(__AVX2__) */
--- a/blosc/bitshuffle-avx2.c
+++ b/blosc/bitshuffle-avx2.c
@@ -18,9 +18,7 @@
 
 
 /* Make sure AVX2 is available for the compilation target and compiler. */
-#if !defined(__AVX2__)
-  #warning AVX2 is not supported by the target architecture/platform and/or this compiler.
-#else
+#if defined(__AVX2__)
 
 #include <immintrin.h>
 
--- a/blosc/bitshuffle-sse2.c
+++ b/blosc/bitshuffle-sse2.c
@@ -17,8 +17,8 @@
 
 /* Make sure SSE2 is available for the compilation target and compiler. */
 #if !defined(__SSE2__)
-  #error SSE2 is not supported by the target architecture/platform and/or this compiler.
-#endif
+  #warning SSE2 is not supported by the target architecture/platform and/or this compiler.
+#else
 
 #include <emmintrin.h>
 
@@ -465,3 +465,5 @@ int64_t blosc_internal_bshuf_untrans_bit_elem_sse2(void* in, void* out, const si
 
     return count;
 }
+
+#endif /* !defined(__SSE2__) */
--- a/blosc/bitshuffle-sse2.c
+++ b/blosc/bitshuffle-sse2.c
@@ -16,9 +16,7 @@
 #include "bitshuffle-sse2.h"
 
 /* Make sure SSE2 is available for the compilation target and compiler. */
-#if !defined(__SSE2__)
-  #warning SSE2 is not supported by the target architecture/platform and/or this compiler.
-#else
+#if defined(__SSE2__)
 
 #include <emmintrin.h>
 
--- a/blosc/shuffle-avx2.c
+++ b/blosc/shuffle-avx2.c
@@ -11,8 +11,8 @@
 
 /* Make sure AVX2 is available for the compilation target and compiler. */
 #if !defined(__AVX2__)
-  #error AVX2 is not supported by the target architecture/platform and/or this compiler.
-#endif
+  #warning AVX2 is not supported by the target architecture/platform and/or this compiler.
+#else
 
 #include <immintrin.h>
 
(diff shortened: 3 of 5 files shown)


t20100:
CI's broken for `CMake / Windows GCC ` but it looks to be the same for recent PRs.

kalvdans:
So if I understand `universal2` correctly, a single invocation of the apple clang compiler preprocess each input file twice, once with `__AVX2__` and `__SSE2__` defined, and once with them undefined (to build arm binary)?

FrancescAlted:
In addition to @kalvdans comment, I'd suggest to use (or plan to adopt) C-Blosc2 for creating universal binaries, as it has native support for NEON in ARM machines.  C-Blosc2 [brings other interesting features](https://www.blosc.org/c-blosc2/c-blosc2.html#new-features-in-c-blosc2), and is backward compatible with C-Blosc1.

FrancescAlted:
> CI's broken for `CMake / Windows GCC ` but it looks to be the same for recent PRs.

Yeah, that's unfortunate and should be fixed somehow.

t20100:
>So if I understand universal2 correctly, a single invocation of the apple clang compiler preprocess each input file twice, once with __AVX2__ and __SSE2__ defined, and once with them undefined (to build arm binary)?

Yes a single call to `clang` with `-arch arm64 -arch x86_64` produces a single `.o` file with the build for both architectures and the macros `__AVX2__`, `__SSE2__` have different values according to the architecture.

> If that is the case, the changes to shuffle.c are unnecessary

Something needs to be done in `shuffle.c` because, unless I missed something with `universal2`,  `SHUFFLE_AVX2_ENABLED` and `SHUFFLE_SSE2_ENABLED` can only be defined once for both architectures.
We want to define them for x86, but in the case of ARM, `blosc_internal_*_avx2` and `blosc_internal_*_sse2` functions do not exists but would then be referred to to fill `impl_avx2` and  `impl_sse2` structures.
An alternative could be to define `blosc_internal_*_avx2` and `blosc_internal_*_sse2` functions even when `__AVX2__`/`__SSE2__` are not defined with a body calling e.g., `abort()`.

t20100:
>  I'd suggest to use (or plan to adopt) C-Blosc2

+1!
But I face this issue for packaging the blosc(1) HDF5 filter for ease of use with `h5py` in the [`hdf5plugin`](https://github.com/silx-kit/hdf5plugin) Python package, and according to https://github.com/Blosc/hdf5-blosc/issues/29 using Blosc-2 should be possible but would produce compressed chunks that are incompatible with previous versions of the filter.

kalvdans:
> in the case of ARM, `blosc_internal_*_avx2` and `blosc_internal_*_sse2` functions do not exists but would then be referred to to fill `impl_avx2` and `impl_sse2` structures.

I would agrue that those structures should not exist on ARM.

t20100:
> I would argue that those structures should not exist on ARM.

Thus the use of `defined(SHUFFLE_AVX2|SSE2_ENABLED) && defined(__AVX2|SSE2__)` in `shuffle.c` since it is where they are defined.

kalvdans:
> Thus the use of `defined(SHUFFLE_[AVX2|SSE2]_ENABLED) && defined(__[AVX2|SSE2]__)` in `shuffle.c` since it is where they are defined.

Got you! I now understand your PR and I'm in favour of the changes.

blosc/bitshuffle-avx2.c
```diff
@@ -19,8 +19,8 @@
 
 /* Make sure AVX2 is available for the compilation target and compiler. */
 #if !defined(__AVX2__)
-  #error AVX2 is not supported by the target architecture/platform and/or this compiler.
-#endif
+  #warning AVX2 is not supported by the target architecture/platform and/or this compiler.
```
kalvdans, line 22: Don't print a warning, it is annoying to have to ignore it and it fails to compile with `-Werror`

kalvdans (changes requested):

blosc/bitshuffle-avx2.c
t20100: I removed them.

kalvdans (approved):

FrancescAlted:
> > I'd suggest to use (or plan to adopt) C-Blosc2
> 
> +1! But I face this issue for packaging the blosc(1) HDF5 filter for ease of use with `h5py` in the [`hdf5plugin`](https://github.com/silx-kit/hdf5plugin) Python package, and according to [Blosc/hdf5-blosc#29](https://github.com/Blosc/hdf5-blosc/issues/29) using Blosc-2 should be possible but would produce compressed chunks that are incompatible with previous versions of the filter.

It occurs to me that it should be relatively easy for C-Blosc to understand chunks from C-Blosc2 (besides the additional 16 bytes in the [Blosc2 header](https://github.com/Blosc/c-blosc2/blob/main/README_CHUNK_FORMAT.rst#header), they are largely compatible).  Would that help your migration plans?

FrancescAlted (approved):

t20100:
`hdf5plugin` is only a way to package HDF5 filters for use with `h5py` and so to ease their installation/usage.
So we'll follow what is done in https://github.com/Blosc/hdf5-blosc ;)

> Would that help your migration plans?

Not sure, or as a transition (but does it worth it? and is it the way to go with HDF5 filters?).
I added more comments here: https://github.com/Blosc/hdf5-blosc/issues/29#issuecomment-1288992875

t20100:
Hi,

Do you plan on tagging a version of `c-blosc` in the near future?
Otherwise I'll reference the merge commit from [`hdf5plugin`](https://github.com/silx-kit/hdf5plugin).

PR #348, opened 2022-10-26
Upgrade internal-complib zstd from 1.5.0 to 1.5.2

mkitti:
This pull requests upgrades zstd from version 1.5.0 to version 1.5.2, the latest released version at https://github.com/facebook/zstd/releases

xref: https://github.com/HDFGroup/hdf5_plugins/pull/44


mkitti:
@FrancescAlted, it would be great if we could incorporate this and do a patch release so that downstream https://github.com/HDFGroup/hdf5_plugins/pull/44 can use the latest released version of zstd. Thanks, Mark.

FrancescAlted:
> @FrancescAlted, it would be great if we could incorporate this and do a patch release so that downstream [HDFGroup/hdf5_plugins#44](https://github.com/HDFGroup/hdf5_plugins/pull/44) can use the latest released version of zstd. Thanks, Mark.

Ok, for using the latest version of any of the internal codecs one can always compile it by defining a cmake symbol [PREFER_EXTERNAL_ZSTD=ON](https://github.com/Blosc/c-blosc#codec-support-with-cmake).

mkitti:
Yes, I only tested it with MSVC locally.

mkitti:
This would work now. I just backported the c-blosc2 changes to cmake.

FrancescAlted:
LGTM.

mkitti:
Thank you, Francesc.

PR #349, opened 2022-11-08
Consistency between blosc and blosc2

DimitriPapadopoulos:
Follow the CMake manual:
https://cmake.org/cmake/help/latest/command/endif.html

* No space before `(`.
* No deprecated `<condition>` argument to `endif()`.
  >The optional `<condition>` argument is supported for backward compatibility only. If used it must be a verbatim repeat of the argument of the opening `if` clause.


FrancescAlted:
LGTM

PR #351, opened 2023-01-10
Fix CMake Processor Identification Logic

jaw:
CMake string comparison would not work with CMake 3.24.2. Thus, the if statement has been updated to use matches rather than a bunch of STREQUAL which seems to work better.


FrancescAlted:
LGTM.  Would you mind providing a similar PR for the C-Blosc2 sibling project?  Thanks!

PR #352, opened 2023-01-10
Fix SSE2/AVX2 build issue

t20100:
This PR fixes a regression introduced in PR #347:

`shuffle.c` uses `__SSE2__` and `__AVX2__` to set `SHUFFLE_USE_SSE2` and `SHUFFLE_USE_AVX2` but those macros where never defined since `shuffle.c` was not built with those flags.

It is similar to https://github.com/Blosc/c-blosc2/pull/436


FrancescAlted:
Thanks @t20100 !

PR #358, opened 2023-02-02
LICENSES/BLOSC.txt → LICENSE.txt

DimitriPapadopoulos:
For consistency with c-blosc2:
https://github.com/Blosc/c-blosc2/commit/c9b6e6c


DimitriPapadopoulos:
Or we could align c-blocs2 on c-blosc, the other way round, if you'd rather prefer that.

FrancescAlted:
I like this way better.  Thanks!

PR #360, opened 2023-02-18
Upgrade internal-complib zstd from 1.5.2 to 1.5.4

DimitriPapadopoulos:
I have extracted the `lib` directory from the Zstandard tarball, and discarded the rest, as was the case for previous versions.


DimitriPapadopoulos:
Perhaps the errors are related to this new paragraph in [zstd/lib/README.md](https://github.com/facebook/zstd/blob/dev/lib/README.md#modular-build)

> - The C compiler macros `ZSTDLIB_VISIBLE`, `ZSTDERRORLIB_VISIBLE` and `ZDICTLIB_VISIBLE` can be overridden to control the visibility of zstd's API. Additionally, `ZSTDLIB_STATIC_API` and `ZDICTLIB_STATIC_API` can be overridden to control the visibility of zstd's static API. Specifically, it can be set to `ZSTDLIB_HIDDEN` to hide the symbols from the shared library. These macros default to `ZSTDLIB_VISIBILITY`, `ZSTDERRORLIB_VSIBILITY`, and `ZDICTLIB_VISIBILITY` if unset, for backwards compatibility with the old macro names.

FrancescAlted:
LGTM

PR #362, opened 2023-02-26
Always use inline assembly for _xgetbv on GCC/Clang

jbms:
This reverts the change to blosc/shuffle.c in
0b8608b5d12ca96d4ac4525b36a62062b2286fd4.

The _xgetbv intrinsic defined by GCC and Clang can only be used when built with support for the xsave target feature.  Therefore, attempting to use it from blosc_get_cpu_features results in a compiler error:

error: inlining failed in call to 'always_inline' '_xgetbv': target specific option mismatch

On most platforms, the intrinsic was not being used because immintrin.h was not getting transitively included in shuffle.c.  However, on mingw64 it does get transitively included in shuffle.c, which leads to the error.

In order to be able to use the intrinsic, it would need to be used from a separate function that is built with xsave support.  That is more complicated than just using the inline assembly version.

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -182,8 +182,6 @@ blosc_internal_cpuidex(int32_t cpuInfo[4], int32_t function_id, int32_t subfunct
 
 #define _XCR_XFEATURE_ENABLED_MASK 0
 
-#if !(defined(_IMMINTRIN_H_INCLUDED) && (BLOSC_GCC_VERSION >= 900))
-
 /* Reads the content of an extended control register.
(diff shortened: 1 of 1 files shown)


FrancescAlted:
I don't actually remember why I did this change back in the day.  Apparently all the tests are passing, so I'm +1 on merging this.

BTW, do you think that C-Blosc2 would be bitten by this too?  Thanks!

jbms:
Thanks.  From a quick look at c-blosc2, yes I think the same issue applies.

PR #366, opened 2023-05-01
No need to support Python 2 any more

DimitriPapadopoulos:
As in https://github.com/Blosc/c-blosc2/pull/468.


FrancescAlted:
Thanks!

PR #370, opened 2023-05-14
Add const where possible

DimitriPapadopoulos:


tests/fuzz/fuzz_compress.c
```diff
@@ -8,7 +8,7 @@ extern "C" {
 #endif
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
```
kalvdans, line 10:
```suggestion
int LLVMFuzzerTestOneInput(const uint8_t *const data, const size_t size) {
```
Add const to the parameters also while you are at it. (These doesn't change the signature, just make them immutable in the function body)

Personally I think it is silly to add const to local variables, if it aids readability then function is too big :D

kalvdans (approved):

tests/fuzz/fuzz_compress.c
DimitriPapadopoulos: The variable is passed by value, so `const` doesn't really help.

Also, the required signature is documented here:
https://llvm.org/docs/LibFuzzer.html#fuzz-target

kalvdans: `const` helps to not inadvertly modify the local copy inside the function body. It doesn't change the function signature.

FrancescAlted:
Thanks @DimitriPapadopoulos

PR #378, opened 2024-03-19
Align LICENSE.txt with c-blosc2

DimitriPapadopoulos:


FrancescAlted:
Thanks @DimitriPapadopoulos .  IIRC there is some dancing around the "The " prefix.  The Jupyter project prefers without the "The ", so it may be a good idea to uniformize and convert "The Blosc Development Team" -> "Blosc Development Team" everywhere.

PR #383, opened 2024-08-11
Internal LZ4 codec updated to 1.10.0

DimitriPapadopoulos:
> This version brings multithreading support to the forefront, harnessing modern multi-core processors to accelerate both compression and decompression processing. It's a good upgrade for users looking to optimize performance in high-throughput environments.


FrancescAlted:
Thanks @DimitriPapadopoulos !

PR #387, opened 2024-11-03
Move getenv out of a function used in multiple threads

### LINKED ISSUE #384: `getenv` used in `blosc_compress_ctx`
From what I can tell `getenv` is used in `blosc_compress_ctx` here:
https://github.com/Blosc/c-blosc/blob/3455c3810279ce709cf02f45beddbb31af418ab6/blosc/blosc.c#L1089

`blosc_compress_ctx` is supposed to be used in multi-threaded contexts; however, `getenv` is not threadsafe:
https://pubs.opengroup.org/onlinepubs/9699919799.2016edition/functions/getenv.html

nhz2:
Fixes #384

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -1069,10 +1069,9 @@ static int initialize_context_compression(struct blosc_context* context,
                           size_t destsize,
                           int32_t compressor,
                           int32_t blocksize,
-                          int32_t numthreads)
+                          int32_t numthreads,
+                          long warnlvl)
 {
-  char *envvar = NULL;
-  int warnlvl = 0;
   /* Set parameters */
   context->compress = 1;
   context->src = (const uint8_t*)src;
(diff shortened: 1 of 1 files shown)


blosc/blosc.c
```diff
@@ -1069,10 +1069,9 @@ static int initialize_context_compression(struct blosc_context* context,
                           size_t destsize,
                           int32_t compressor,
                           int32_t blocksize,
-                          int32_t numthreads)
+                          int32_t numthreads,
+                          long warnlvl)
```
kalvdans, line 1073: warnlvl is just a small integer, so use `int`. You can still use strtol.

kalvdans (approved):
nice! (i don't have write permission though)

nhz2:
Thank you for the review

FrancescAlted:
Thanks @nhz2 !

PR #388, opened 2024-12-06
Fix typo found by codespell

DimitriPapadopoulos:


kalvdans (approved):
codespell finding incorrect names? Haenel occurs on more times in the file with Valentin so I agree it is a misspelling.

PR #390, opened 2024-12-11
Clamp `destsize` to avoid overflow

### LINKED ISSUE #159: Blosc 1.7.0 compression uses more than uncompressed_size + BLOSC_MAX_OVERHEAD
I have observed the following, but I am unsure, if this can be called an Issue.

Using a comp_buffer_size much larger than uncompressed_size and executing 

int compressedSize = blosc_compress(compress_level,shuffel,block_size,
                                            uncompressed_size,uncompressed_buffer,
                                            comp_buffer,comp_buffer_size);

the compressedSize can be larger than uncompressed_size + BLOSC_MAX_OVERHEAD, but will always be smaller than comp_buffer_size. For me, this was not clear from the documentation, I had the assumption, that compressedSize <= uncompressed_size + BLOSC_MAX_OVERHEAD would always hold true.

This can be cured by working with 

int compressedSize = blosc_compress(compress_level,shuffel,block_size,
                                            uncompressed_size,uncompressed_buffer,
                                            comp_buffer,uncompressed_size + BLOSC_MAX_OVERHEAD );

Is there a performancechange between the two? Is this expected behavior?

### LINKED ISSUE #389: Integer overflow error for `destsize` argument to `blosc_compress_ctx`
Here is a MWE:

```c
#include <stdio.h>
#include <stdint.h>
#include <blosc.h>

#define SRC_SIZE 1024
#define DST_SIZE 4294967296ULL

int main(){
  
  /* Allocate Input and output data byte buffers*/
  uint8_t *src = malloc(SRC_SIZE);
  uint8_t *dst = malloc(DST_SIZE);
  /* Check if the allocation was successful*/
  if(src == NULL || dst == NULL){
    printf("Memory allocation failed\n");
    return 1;
  }

  /* Fill the input data buffer  with random bytes*/
  srand(1234);
  for(int i=0;i<SRC_SIZE;i++){
    src[i] = (uint8_t)rand();
  }

  int csize = blosc_compress_ctx(5, 1, 1,
                            SRC_SIZE, src, dst, DST_SIZE,
                            "lz4", 0, 1);
  printf("Compression Returned: %d\n", csize);
  return 0;
}
```

Running this on a 64-bit system I get:

```
Compression Returned: 0
```
But compression should succeed.

The overflow is happening at:
https://github.com/Blosc/c-blosc/blob/dcf6813d46f9fcbf7af4395b5536f886dfc76dd0/blosc/blosc.c#L1080

There are some checks on `destsize` being too small, but from what I can tell, there are no checks on it being too large.
https://github.com/Blosc/c-blosc/blob/dcf6813d46f9fcbf7af4395b5536f886dfc76dd0/blosc/blosc.c#L1096-L1102

One option to fix this would be to clamp `destsize` to be at most `sourcesize + BLOSC_MAX_OVERHEAD`, this would also fix #159

nhz2:
I am also suppressing more warning printing if `warnlvl` is zero.

Fixes #389 and fixes #159

### DIFF
--- a/blosc/blosc.c
+++ b/blosc/blosc.c
@@ -1072,20 +1072,7 @@ static int initialize_context_compression(struct blosc_context* context,
                           int32_t numthreads,
                           int warnlvl)
 {
-  /* Set parameters */
-  context->compress = 1;
-  context->src = (const uint8_t*)src;
-  context->dest = (uint8_t *)(dest);
-  context->num_output_bytes = 0;
-  context->destsize = (int32_t)destsize;
-  context->sourcesize = sourcesize;
-  context->typesize = typesize;
-  context->compcode = compressor;
-  context->numthreads = numthreads;
-  context->end_threads = 0;
-  context->clevel = clevel;
-
-  /* Check buffer size limits */
+  /* Check buffer size limits and clamp destsize */
   if (sourcesize > BLOSC_MAX_BUFFERSIZE) {
     if (warnlvl > 0) {
       fprintf(stderr, "Input buffer size cannot exceed %d bytes\n",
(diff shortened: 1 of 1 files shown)


kalvdans (approved):
Beautiful! LGTM (but I have no merge rights)

FrancescAlted:
Thanks @nhz2 !

PR #392, opened 2025-04-17
Drop a slightly outdated type def for boolean type

weiznich:
This cases compilation errors with gcc 15 with the following error message:

c-blosc/blosc/shuffle.c:26:15: error: 'bool' cannot be defined via 'typedef'
      | typedef _Bool bool;
      |               ^~~~
c-blosc/blosc/shuffle.c:26:15: note: 'bool' is a keyword with '-std=c23' onwards

See https://gcc.gnu.org/gcc-15/porting_to.html for details

I've choosen to remove the typedef as it seems to be unused in this file.

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -20,16 +20,6 @@
 #include <pthread.h>
 #endif
 
-/* Visual Studio < 2013 does not have stdbool.h so here it is a replacement: */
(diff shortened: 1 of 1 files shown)


LDeakin (approved):

tacaswell:
closes #393

tacaswell (approved):

tacaswell:
The usage was dropped in a8db55b0040e1df72422f33ce1100fa6a66c773c (2016-07-29).

attn @FrancescAlted

PR #396, opened 2025-08-04
Rename windows pthread functions to blosc_pthread_{xyz}

barakugav:
This PR changes a few things:
- Add `threading.h`, exposing a unified threading API for UNIX and Windows
- Rename pthread functions to `blosc_pthread_{xyz}`, by (1) `#define blosc_pthread_{xyz} pthread_{xyz}` on UNIX and (2) actually implementing the function in `threading.c` on Windows. This is done to avoid exposing pthread_{xyz} symbols, as they may clash with other libraries.
- Add threading.c to the sources list on windows even if the threading library is found. Previously, the source was added in CMakeLists.txt only in specific cases, but was also included by `#include "win32/pthread.c"` on any windows build, now we are just more explicit about it (also including a .c file is a bit ugly)

The approach of defining `blosc_pthread_{xyz}` as either alias to the regular pthread API or to the windows API is common, for example it is used by the zstd library, see `internal-complibs/zstd-1.5.6/common/threading.h`

This PR is a port of a similar change in blosc2 https://github.com/Blosc/c-blosc2/pull/674


blosc/blosc.c
```diff
@@ -59,7 +59,6 @@
 /* Include the win32/pthread.h library for all the Windows builds. See #224. */
```
kalvdans, line 59: The comment explains why the .c file is included, please remove the comment too.

FrancescAlted:
This PR is not passing tests for MSVC (maybe this was the reason for using the code as is).

blosc/blosc.c
barakugav: done

FrancescAlted:
Sorry, but we don't have resources for continuing the development of C-Blosc 1.x, and all the action is happening in C-Blosc2.  We only accept pure maintenance/security patches for 1.x series.  If C-Blosc2 does not fit your needs, we will be happy to hear about that.

FrancescAlted:
Closing.

PR #399, opened 2025-11-15
Set -std=gnu99 for gcc on all architectures

kxxt:
Previously the cmake file only sets `-std=gnu99` when it is on x86(-64), which causes compilation failure on other architectures when gcc is used.

GCC currently defaults to `-std=gnu23`, which would fail the build with:

    blosc/shuffle.c:26:15: error: ‘bool’ cannot be defined via ‘typedef’

This patch fixes it by setting `-std=gnu99` for gcc on all architectures.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,6 +207,11 @@ if(APPLE)
   set(CMAKE_MACOSX_RPATH TRUE)
 endif()
 
+if(CMAKE_C_COMPILER_ID STREQUAL GNU)
+    # We need C99 (GNU99 more exactly)
+    SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
+endif()
+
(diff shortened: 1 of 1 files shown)


kalvdans:
Why first check if the compiler is gcc? If our code is written in the gnu99 dialect, it should be passed to any compiler. Clang also supports compiling gnu99 code.

kxxt:
> Why first check if the compiler is gcc? If our code is written in the gnu99 dialect, it should be passed to any compiler.

The original code sets it to `gnu99` only for the GCC compiler. So I kepted that.

> Clang also supports compiling gnu99 code.

That's nice! But I suppose msvc don't support it?

kalvdans:
> That's nice! But I suppose msvc don't support it?

I doubt. But how can it then compile the code?

kalvdans (approved):

PR #403, opened 2026-06-11
Add generic RISC-V build path

carlosqwqqwq:
## Why

c-blosc reported riscv targets as unknown processors and left its generic fallback path noisy in `shuffle.c`, even though the generic C implementation was otherwise usable. This patch makes the baseline RISC-V build path explicit while preserving the existing generic implementation.

## What changed

- Recognize `riscv*` targets explicitly in the top-level CMake processor selection instead of reporting them as unknown processors.
- Keep RISC-V builds on the existing generic shuffle/bitshuffle path without enabling SSE2 or AVX2 sources or flags.
- Let `blosc/shuffle.c` take a quiet generic fallback on RISC-V instead of emitting the non-x86 hardware-acceleration warning.
- Preserve the existing x86 acceleration logic and only adjust the conservative non-x86 build path.

## Verification

- Ran native CMake configuration with Ninja on the local Windows host.
- Built the native tree successfully with `cmake --build build-codex-native --parallel 4`.
- Ran native test verification with `ctest --test-dir build-codex-native --output-on-failure`; all `1639/1639` tests passed.
- Ran simulated `riscv64` CMake configuration with `CMAKE_SYSTEM_NAME=Linux`, `CMAKE_SYSTEM_PROCESSOR=riscv64`, `CMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY`, `BUILD_SHARED=OFF`, `BUILD_STATIC=ON`, `BUILD_TESTS=OFF`, `BUILD_FUZZERS=OFF`, `BUILD_BENCHMARKS=OFF`, and forced `__riscv=1` / `__riscv_xlen=64` macros because no real RISC-V toolchain is available on the host.
- Built the simulated `riscv64` static library successfully with `cmake --build build-codex-riscv --parallel 4`.
- Checked `build-codex-riscv/build.ninja`: it contains `shuffle-generic.c`, `bitshuffle-generic.c`, `shuffle.c`, `ZSTD_DISABLE_ASM`, `__riscv=1`, and `__riscv_xlen=64`.
- Confirmed `build-codex-riscv/build.ninja` does not contain `shuffle-sse2.c`, `bitshuffle-sse2.c`, `shuffle-avx2.c`, `bitshuffle-avx2.c`, `-msse2`, `-mavx2`, or `huf_decompress_amd64.S`.

## Notes

This is a conservative portability patch. It does not add RVV kernels or any RISC-V-specific performance optimizations. Verification covers native build and test stability plus simulated RISC-V target selection and source/flag exclusion checks, but not a real riscv64 cross-toolchain build or hardware run.


blosc/shuffle.c
kalvdans: Remove the noisy compiler output for all non-accellerated architectures, as cmake has already warned about it.

carlosqwqqwq: Thank you for pointing this out. I removed the noisy generic fallback warning for all non-accelerated architectures and rechecked the native tests plus the simulated riscv64 build locally; I’m still learning RISC-V, so I really appreciate your guidance.

